PR #2, opened 2014-12-14
Correct KSFO Airport Designator

pcrumm:


zlsa:
I'll merge this now.

It was KDBG for "debug airport" and I copy-pasted the airport; however, the layout isn't correct. I'll add a handful of real-life airports soon.

PR #8, opened 2014-12-14
Fix confusing wording in tutorial

albertyw:
The tutorial uses the word "arrow" several times to mean several different things - next step, speed up the simulation, and the arrow button on your keyboard.  This is a bit confusing.  I don't know what can be done without changing the image assets, but hopefully this PR can start it off.  

Nice job though.


zlsa:
There are no image assets except for the four icons to the right of the command box - everything else is drawn manually on a canvas or in the DOM.

Issue #3, opened 2014-12-14
Real life airport layouts


zlsa:
There's only a demo airport layout included; it would be nice to include multiple real-life airports. The format is a bit arcane, though.

pcrumm:
Curious about your thoughts on this. I started playing around with SFO:
https://github.com/pcrumm/atc/blob/add-sfo/assets/airports/ksfo.json

I added the ILS/LOC approach 28L and 28R. There are no designated radio-assisted approaches for the other runways, so we'll need to make up fixes if we want a more symmetrical layout. You've spent more time thinking about the game than I have--better to leave it as is (and see operation like SFO really does), or to add some fixes on the other sides of the airport to make things easier? I'll probably work out the math for a few other local fix positions before I PR, but there won't be anywhere near as many, obviously.

psycho202:
Currently working on JFK Intl airport.
Adding the fixes atm.

https://github.com/psycho202/atc/blob/kjfk/assets/airports/kjfk.json

stijnsanders:
is there an online repository of waypoints? I checked http://www.faa.gov/ and http://www.eurocontrol.int/ but I guess a search from the homepage won't hand them to me
(but with just a little scratching: https://nfdc.faa.gov/xwiki/bin/view/NFDC/Browse+Aeronautical+Data )

psycho202:
Dear Stijn: For the US airports, both me (who made KJFK) and arithmetric (who made KMSP) used the airport diagrams found on Flightaware.com.

Using KMSP as example: http://nl.flightaware.com/resources/airport/KMSP/IAP/ILS+RWY+30L%28CAT+II%29

There you can find all the waypoints in the Approach and Depart procedures.

pcrumm:
These are also available from airnav:
http://www.airnav.com/airport/KSFO

In particular, you'll want to look for ILS/LOC approaches and VOR approaches, and scan over departure procedures, though you'll generally find that departure fixes are far enough away from the airport that they aren't really includable here without some fudging.

stijnsanders:
ooh look what I found! would this make an interesting layout?
http://www.belgocontrol.be/website/eaip/eAIP_Main/html/eAIP/EB-AD-2.EBBR-en-GB.html

tomasfg:
If you want some info about spanish airports you can have a look at this page:

http://www.enaire.es/csee/Satellite/navegacion-aerea/en/Page/1078418725163/?other=1083158950596#ancla3

[...]

Issue #9, opened 2014-12-14
No more aeroplanes.


popey456963:
Am I missing something, or after a certain period (less than 15 minutes), do no more planes come?

http://i.imgur.com/MUW6K3c.png (just, ignore the score.)

Running chrome, windows 7.

zlsa:
Is there anything printed out in the console after it loads (and about 10 aircraft are created)? The updated version is 2.1.1, upper-right corner while loading.

popey456963:
I get this message:

http://pastebin.com/WX7CdHr2

Which seems that creating aeroplanes is definitely working.  I will land/fly all planes and see if more are generated.

jvt:
Just like to add in my log output here, maybe it'll help: http://pastebin.com/0Yy45gcJ

It seems like the plane generation is working until it reaches a certain number at which point it will stop generating any more.

popey456963:
It seems as if it isn't a problem generating the aeroplanes, it's actually a problem calling the function which generates the aeroplanes.  
Where is the "aircraft_new()" function called?

psycho202:
Popey: the aircraft_new function is being called in airport.js. 
At lines 256 to 260 for departures and at lines 296 to 303 for arrivals.

Both of these are being called in addAircraft, at line 227 in airport.js.

joallard:
+1 on Firefox Aurora 36

joallard:
I'm debugging around a little. As I understand it: game initally adds some departure, and that function will set up timeouts to add some more departure, based on the `timeout` parameter it receives. That parameter would be based on the big timeout array the game receives as options. But in the array, the timeout parameter the function will receive is set to `false`, hence no timeout. ("timeout" comes up a lot in this comment, but that is the nomenclature used in the code)

zlsa:
Can reproduce, fix in progress.

I've never played it for more than about ten minutes so I never encountered this bug...

joallard:
I just noticed it wasn't a big setting array, but rather just the timeouts added by the game to the queue, as async tasks. The `false` param on the call at airport.js:233 (Airport#addAircraft()) looks suspicious.

zlsa:
For testing, you can run the command `timewarp 100` without a callsign (warning: high numbers will slow down your computer); to return to normal speed, type `timewarp 1`. This bug should be gone now.

joallard:
Fix confirmed. However, the spawn rate is now too high. My list keeps growing and growing.

jvt:
[...]

PR #21, opened 2014-12-15
Re-vamp SFO to Reflect Reality

pcrumm:
This is a step forward for #3. The default SFO is a single runway, which is kind of lame. This duplicates the "real" SFO, with realistic winds (taken from earlier today), actual runway arrangement, and actual approach fixes for ILS runways.

Could use a little more love (and some more local fixes), but it's better than what's there.


pcrumm:
This might need a revisiting after #9, as it copies KDBG's spawn rates which now seem to be a little ridiculous.

PR #25, opened 2014-12-15
Tab completion

garcg:
Completion of callsigns and commands with the `tab` key.

Repeatedly pressing `tab` will cycle through the current matches for the aircraft or command. In case of an aircraft, it is highlighted on the map/sidebar accordingly. Pressing `shift+tab` cycles backward.

Currently an incomplete callsign/command is valid as long as it is unambiguous. This PR _does not_ change that, it just adds additional completion and cycling through ambiguous matches.

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -187,37 +187,36 @@ var Aircraft=Fiber.extend(function() {
       }
       return airline_get(this.airline).callsign.name + " " + callsign.toUpperCase() + heavy;
     },
+    COMMANDS: [
+      "turn",
+      "heading",
+
+      "altitude",
+      "climb",
+      "clear",
+      "descend",
+
+      "speed",
+      "slow",
+
+      "hold",
+      "circle",
+
+      "fix",
+
+      "wait",
+      "taxi",
+
+      "takeoff",
+
+      "land",
+
+      "abort",
+
+      "debug"
+    ],
     runCommand: function(command) {
       command = command.toLowerCase();
-      var COMMANDS = [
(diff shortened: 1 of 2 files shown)


assets/scripts/aircraft.js
```diff
@@ -187,37 +187,36 @@ var Aircraft=Fiber.extend(function() {
...
-      var COMMANDS = [
-        "turn",
-        "heading",
-
-        "altitude",
-        "climb",
-        "clear",
-        "descend",
+    COMMANDS: [
+      "turn",
```
garcg, line 191: The only actual change to aircraft.js is moving `COMMANDS` outside `runCommand` so it is accessible from input.js.
The diff shows a lot changes, but it is just that.

psycho202:
Is the tab completing of commands necessary? A short input like "n641mf h 0" is already enough to change heading to 0°. Almost all other commands are possible like that too.

boreq:
I totally support the idea of tab completion, especially of the call signs.

garcg:
The short commands still work. "n641mf h 0"  is still valid with this PR.

I don't think completing commands is _absolutely_ necessary, just a bit nicer. But perhaps it is a personal thing, out of habit.

Now, completing and cycling through call signs, I believe that is an improvement.

psycho202:
Yeah, the tabbing through callsigns is a major advantage.
With the current atc kdbg airport spawn rates, it's a lot easier that way., instead of having to scroll through the list, which also kinda bugs out whenever it needs to scroll.

zlsa:
Works very well, thanks @garcg!

PR #32, opened 2014-12-15
Adding John F Kennedy International Airport (KJFK)

psycho202:
Adding John F Kennedy International Airport (KJFK).
Scale of airport vs onscreen scale is currently out of proportion. 

Wind is updated to dec 15 16:00 angle and speed.

### DIFF
--- a/assets/airports/kjfk.json
+++ b/assets/airports/kjfk.json
@@ -3,11 +3,16 @@
   "level": "hard",
   "icao": "KJFK",
   "wind": {
-    "angle": 315,
+    "angle": 225,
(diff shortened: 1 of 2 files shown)


zlsa:
The scale of all distances is KM, so your airport is ~10x larger than it should be (since the longest runway is 4km, but the airport shows it being 20km instead. (I'll fix it now; if you fix it we'll just have to merge changes). I do plan on documenting the format soon, but I'm rewriting the arrival spawn algorithm right now and want to finish that first as the airport format encodes arrival and departure timing.

PR #38, opened 2014-12-16
concise String.split in aircraft.js

niklasfi:
replace complicated tokenisation logic for commands with single functionally equivalent function call

For details for String.prototype.split see its explanation in Mozilla Developer
Network [1].


zlsa:
Thanks - the tokenization was a relic from before I rewrote the command system.

PR #40, opened 2014-12-16
Fix A388 designation, add e-series jets and B762, add airlines AAL and AWE

daviessm:


zlsa:
FYI, until the airlines are added to the airports they frequent, they won't ever appear. Watch the wiki - I'll be adding the format descriptions.

daviessm:
I know, I was adding them so people could extend airports if they wished. I might do some more later. :-)

PR #53, opened 2014-12-18
Improvements to KJFK

groeneman:
- Runway numbers are more readable
- Add radio callsign
- Fix incorrectly swapped runways 13/31


zlsa:
I shouldn't edit when tired... thanks, @Diastro.

PR #57, opened 2014-12-18
 fix bug where runway was still represent after a landing was cancelled

niklasfi:
to reproduce: tell plane to land. "no ILS" is displayed. Tell plane to
navigate to a fix "no ILS" is still visible, although it should be gone.


Diastro:
Im not familiar enough with this piece of the code to merge so I'll let @zlsa do it.

That being said, i've tried to repro the bug you are mentioning but I can't seem to repro it with the current build.

If I do : plane a land 34R (no ILS is shown) then I do plane a direct ABOFA no ILS disappears and the altitude is coming back.... Am I missing something?

niklasfi:
@daviessm you may be right. I am a little embarassed now :(. On the other hand, this does fix a bug. Try the same with heading, and you will see what the problem is. `plane land 36R`, `plane heading 0`

Diastro:
@niklasfi Yup you're right one that one! (No need to be embarrassed your help is very much appreciated :+1: )

I'll leave @zlsa do the merge on that one.
Thanks

zlsa:
I'll hand-merge this soon. Thanks @niklasfi and @Diastro.

Diastro:
Also, issuing a command like : PlaneA direct (without any fix) will make the sim crash.

Info : Uncaught TypeError: Cannot read property '0' of null, aircraft.js:861

@zlsa agian since i'm not yet familiar with this piece of code you might have to take that one...

zlsa:
@Diastro, fixed that as well.

PR #70, opened 2014-12-22
ebbr.json doesn't exist, causing the load to fail

quink:
ebbr.json is missing, causing http://zlsa.github.io/atc/ to be stuck on:

Failed to load resource: the server responded with a status of 404 (Not Found)
modules.js:125 [ WARN  ] Failed to get assets/airports/ebbr.json: 404

With retries every few seconds.

Either check in ebbr.json or the commit here will remove the reference to it.


zlsa:
Sorry about that; I was testing a new layout and forgot to remove it.

PR #77, opened 2015-01-20
added "to" as a shortcut for "takeoff"

larsen0815:
I don´t know if "to" should also be added to

```
var DEFERRED_COMMANDS = ["takeoff"];
```

What is the purpose of this?
It is pushed to "deferred" afterwards, passed to "run()", passed to "runTakeoff()" but I don´t see it being used there.

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -269,6 +269,7 @@ var Aircraft=Fiber.extend(function() {
       "direct",
 
       "takeoff",
+      "to",
 
       "wait",
       "taxi",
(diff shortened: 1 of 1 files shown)


zlsa:
`deferred` commands are run after all other commands in the current chain; for example:

```
BAW123 climb 250 turn right 090
```

would run the following, in this order:

```
climb turn
```

However, the command

```
BAW123 takeoff climb 250
```

would run in this order:

```
climb takeoff
```

because `takeoff` needs an altitude set (with the `climb` command), and I often typed `takeoff climb 10` instead of `climb 10 takeoff`. 

To answer your question, since `deferred` is generated before the commands are shortened (i.e. `tax` is the same as `taxi`), `to` does have to be added to `DEFERRED_COMMANDS`.

PR #79, opened 2015-01-25
ESC to reset input

larsen0815:
This deletes all text from the input box when you press ESC.


Diastro:
Good idea @larsen0815 . Thanks

PR #81, opened 2015-01-27
Improve speech-output

larsen0815:


Diastro:
Thanks again @larsen0815 !

PR #83, opened 2015-02-11
Add fixes for KSFO RW 1 and 10

satyap:
Not all fixes. I'm a little lost as to where to find them. I just added a few to help noobs like me line up the aircraft.


satyap:
I was adding the fixes against an older commit. The HEAD of gh-pages has the runways shifted, so all the fixes are in a different place relative to the runways.

PR #94, opened 2015-04-17
Fix intermittent "undefined" error

dhleong:
Ran into this the other day. I haven't looked too closely at the source yet, but just glancing at the area this seems to be the correct thing.


zlsa:
Yep, that's a bug. Thanks!

PR #95, opened 2015-04-21
EHAM added

ufkes:
Amsterdam Schiphol International Airport added.


Diastro:
Nice. Thanks for helping!

PR #96, opened 2015-04-23
Rework EHAM and added KLM, KLC

ufkes:
- reworked EHAM coordinates to take proper longitude calculations into account
- added KLM and KLM Cityhopper airlines for operations on EHAM
- increased EHAM ILS ranges to be more realistic (incidentally makes landings easier)
- reworked EHAM arrivals and departure tables to include KLM/KLC, add more diversity to operating airlines, use realistic approach angles, and general increase in difficulty.


Diastro:
Looking good! Thanks for helping!

zlsa:
Thanks @Diastro!

Diastro:
Sorry about the merge/revert. I added your change but when I tried everything EHAM was broken. I ended up re-merging your change since I figured that this is not the issue (everything work when I try in incognito).
I'll have to investigate but i'm guessing the problem I'm getting is coming from the local storage...

ufkes:
That's too bad, it does sound like a local storage problem. Let me know if you need me for anything.

PR #99, opened 2015-04-27
Voice input

dhleong:
Hello! I've been playing with this for a while and thought I'd go ahead and open a PR to get some feedback. I'm not sure it's quite ready to be merged, but it's fairly usable at any rate. 

This adds another option next to "speech" that supports voice command input. I don't have a whole lot of experience with ATC but I tried to match the readbacks and stuff so that saying the commands felt natural. I'm sure however that there could be more flexibility to support more accurate spoken commands. I tried to add "ducking" so that the "speech" and "voice" modules would respect each other. `speech_say` requests will queue up while you're speaking, and voice recognition will stop while `speech_say` is running. While potentially annoying, this does seem to add a bit of realism :) On the other hand, it would probably be nice to be able to reduce the frequency of arrivals when using this mode, since things quickly get unmanageable due to the delay caused by Chrome deciding whether or not you're done speaking.

One thing you'll notice is that there's a lot of comparing and filtering going on. The speech recognition provided by Chrome is apparently more suited to dictation, and while the spec calls for passing in grammar files, it seems that feature is not implemented by anyone. So, callsign recognition in particular has many layers of fallbacks to try to figure out which plane you might have actually meant.

Included is `test/voice.js`, a unit test file intended to be run with node.js. I just got so many weird results from Chrome's speech-to-text that I thought it would be good to document the cases to make sure algorithm tweaks didn't break recognition of other cases. 

To run the test:

```
npm install -g mocha
npm install chai
cd test
mocha voice.js
```

One final thing to note is that Chrome will constantly ask you for permission to use the microphone unless this you're running on https. Github pages seems to support that, which is good, and it's easy enough to set up a self-signed cert for local testing. I used the server provided by `npm install -g http-server`  and run with `hs -S`.

Let me know what you think!

### DIFF
--- a/assets/scripts/game.js
+++ b/assets/scripts/game.js
@@ -90,6 +90,7 @@ function game_pause() {
   $(".pause-toggle").addClass("active");
   $(".pause-toggle").attr("title", "Resume simulation");
   $("html").addClass("paused");
+  voice_stop();
 }
 
 function game_unpause() {
@@ -97,6 +98,7 @@ function game_unpause() {
   $(".pause-toggle").removeClass("active");
   $(".pause-toggle").attr("title", "Pause simulation");
   $("html").removeClass("paused");
+  voice_start();
 }
 
 function game_pause_toggle() {
--- a/assets/scripts/modules.js
+++ b/assets/scripts/modules.js
@@ -6,6 +6,8 @@ var MODULES=[
   "-util",
   "-animation",
 
+  "voice",
+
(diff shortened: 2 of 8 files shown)


zlsa:
Wow, thanks! This is (rather obviously) a big change though, so it'll take a while to go through and make sure it doesn't break anything.

erikquinn:

Closing this PR.

PR #100, opened 2015-05-03
Allow specifying the Voice for speech utterances

dhleong:
Aircraft pick a random voice on creation for a more
colorful experience. Future work could look at all the
international voices offered by Chrome and just
blacklist voices (like "Bubbles," "Whisper," or "Zarvox")
that are unrealistic, creepy, or otherwise break
immersion. See the TODO in speech.js

This is a more digestable PR while we wade through the voice input one :)

Another avenue for future work not mentioned in the commit message is tweaking the `rate` and `pitch` parameters. Some voices are slower or faster than others (Amelie is quite fast!), but generally they have a similar clip, and we could perhaps speed up the rate to reduce audio congestion. Pitch could be randomized for more aural variety. (Rate could as well, for that matter)

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -180,6 +180,9 @@ var Aircraft=Fiber.extend(function() {
 
       };
 
+      // TODO rate, pitch
+      this.voice = speech_pick_voice();
+
       this.parse(options);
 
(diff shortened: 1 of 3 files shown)


erikquinn:

Closing this PR.

Issue #93, opened 2015-02-26
Visual ILS Lock (suggestion)


xenocide0:
I have been trying to come up with a good looking / minimal visual ILS lock indicator:

Anyone else have any ideas? 

![atc_mockup](https://cloud.githubusercontent.com/assets/4967846/6401045/3e19ae9e-bdb2-11e4-944f-856d7397ca3e.jpg)

zlsa:
That looks very good! I like the main one and alt 2; I think 1 and 3 are too complicated while 4 has a square which IMO doesn't fit with the theme. Do you have a branch with this written?

xenocide0:
no, was just trying out ideas in photoshop.

Yea, I liked those the best also, leaning towards alt #2 as it seams the cleanest.

And here is another one:

![ils_lock](https://cloud.githubusercontent.com/assets/4967846/6401312/651cf17a-bdb4-11e4-9ce8-b4779c0f19b8.jpg)

zlsa:
I can pretty easily implement this if we can pick one style. Do you prefer the main one or alt 2?

xenocide0:
main one.    ( Looking at alt#2 now, its sort of looking like a mag glass... )

zlsa:
I'm refactoring the infobox drawing code at the same time so this may take a day or two.

zlsa:
... or longer. Sorry about the huge delay, but real life got in the way (as it often does).

Diastro:
"Real life" is there such a thing! :P

On Tue, May 12, 2015 at 1:30 PM, Jon Ross notifications@github.com wrote:

> ... or longer. Sorry about the huge delay, but real life got in the way
> (as it often does).
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/zlsa/atc/issues/93#issuecomment-101411308.

## 

_David Albertson_

xenocide0:
Damn you real life!

Seriously though, no worries.

xenocide0:
I got the graphics portion working for it, just haven't dug through the code to find where to see if a plane has ILS enabled or locked.

zlsa:
If the aircraft's `mode == "landing"`, then ILS is locked. Thanks for working on it.

xenocide0:
Got a working local branch of it, but not sure how to get it pushed up to GetHub.

avdg:
@xenocide0 Does this help? https://help.github.com/articles/set-up-git/

Maverick283:
[...]

PR #107, opened 2015-06-03
working ILS Lock indicator

xenocide0:
working version of the ILS lock indicator as discussed in: https://github.com/zlsa/atc/issues/93


zlsa:
Very nice, thanks!

Sorry about not being able to do this myself -- I've just been very busy lately.

xenocide0:
No problem, I actually learned a lot about canvas doing this, it was a lot of fun!  Haven't ever worked with it before.

PR #113, opened 2015-06-17
Airport URL-load

AlexMcArrow:
This functionality gives you the ability to save in localStorage pair ICAO-URL. And subsequently download the airport map with with the other server.

Function list:
- airporturl_set(icao, url) - for set new airport by ICAO&url (ex: `airporturl_set('kjfk','https://raw.githubusercontent.com/zlsa/atc/ec82a677508f0941f98ae2f04dda350381f98433/assets/airports/kjfk.json')` in console )
- airporturl_del - for delete airport by ICAO
- airporturl_init - on INIT, preload file by url and create new airport
- airporturl_load - function for load `airport.json` from other server.

Need:
- create ui for work with functions

### DIFF
--- a/assets/scripts/airporturl.js
+++ b/assets/scripts/airporturl.js
@@ -1,0 +1,36 @@
+function airporturl_init() {
+    console.log('airporturl_init');
+    if (localStorage['airporturl'] === undefined) {
+        localStorage['airporturl'] = JSON.stringify({});
+    }
+    var STORAGE = JSON.parse(localStorage['airporturl']);
+    for (var icao in STORAGE) {
+        airporturl_load(STORAGE[icao]);
+    }
+}
+
+function airporturl_set(icao, url) {
+    var STORAGE = JSON.parse(localStorage['airporturl']);
+    STORAGE[icao] = url;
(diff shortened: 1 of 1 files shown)


zlsa:
Is this to allow external airport JSON files to be dynamically loaded?

AlexMcArrow:
yes, like other js or css files (if server not have ORIGIN-block)
You can try exec functions from console for test

glangford:
My opinion only - it doesn't make sense to have outside sources of airport information when the format is custom (and still changing). If there are additional airport definitions, why not contribute them to this repository?

AlexMcArrow:
Add the repository can and should be.
I just thought thereby expanding opportunities.
The decision is yours.

glangford:
Sorry, I don't see what opportunities are expanded. We have a representative set of airports already, and this mechanism only seems to introduce risk to our user community. While I am sure it is not your intention, it could be used to deliver unknown and possibly inappropriate content.

erikquinn:
This can probably be just as easily achieved by forking and adding your own airport. Since no action has been taken on it in over a year, I'll close this for now unless there is a significant interest in it.

PR #118, opened 2015-07-09
Calculate aircraft turn rate based on speed

### LINKED ISSUE #88: Turn radius discrepancy (two craft, same speed, same alt)
It's not as obvious in this picture but it is glaring during game play. 
turn-radius-discrepancy

Repro: Line up two craft on the same heading in close proximity to each other. Put the first craft into a holding pattern, quickly put the second one in to the same holding pattern.

Thanks again for this great code. Super fun game!

tedrek:
This implements a [standard rate turn](https://en.wikipedia.org/wiki/Standard_rate_turn) and would fix #88.  It simulates an aircraft turning at either 3 degrees/second or 25 degrees of bank, whichever is the lower turn rate.

The main thing I find with this change is that the different aircraft models are no longer noticeable, except for the difference between jets and general aviation common speeds.  I find it an acceptable trade-off in terms of increased realism and ability to handle quantity of traffic, but it does give a different character to the game which may not be desirable.

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -1015,7 +1015,11 @@ var Aircraft=Fiber.extend(function() {
         // TURNING
 
         if(this.altitude > 10) {
-          var turn_amount = this.model.rate.turn * game_delta();
+          // Perform standard turns 3 deg/s or 25 deg bank, whichever
+          // requires less bank angle.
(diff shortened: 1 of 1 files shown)


zlsa:
I'll merge this as it's more realistic and isn't a negative change.

PR #119, opened 2015-07-18
Remove duplicate aircraft movement

tedrek:
Aircraft are moving twice the speed which they should be based on assigned airspeeds in knots.  This looks to be unintentional and is fixed by this patch.

Given the difference in gameplay, I'm wondering if this should be offset somehow.  The immediate thought is defaulting to a 2x warp so real time movement ends up approximately the same.


zlsa:
Well shit... I'm not sure what should be done here. This is obviously a bug and my fault but the gameplay will be much simpler... it might be good anyway since there were lots of people complaining about it being too fast. I won't merge it just yet.

tedrek:
A solution which works for me is to set the base aircraft generation rate to 1 thereby doubling the number of aircraft in the game.  The end result is the same number of aircraft through a given point in the same amount of real time.

zlsa:
Sounds good.

Issue #108, opened 2015-06-05
Select traffic loads


sixeyeco:
<h4>A way to dictate the traffic load</h4>

I like to be challenged. If I load the game and there are only a few planes I  will reload it until there is as close to a full board as possible. I already use the speed control to increase the pace of the game so that really doesn't address it. Sometimes I have to load twenty or more times to get a heavy load. Not a big deal, I enjoy the game enough to do it. But if it's relatively easy to implement it would be nice.

So, would it be possible to dictate the initial load (e.g. lite, med, heavy, insane, 9/11 terrorist attack type traffic, etc.)?

zlsa:
It is now, thanks to @tedrek: #120

PR #120, opened 2015-07-19
Add rate command

### LINKED ISSUE #74: Setting for rate of aircraft generation
It'd be great to have a setting that allows users to vary the rate at which new aircraft are generated.

tedrek:
Allow adjusting the rate of aircraft spawning.  At this time
the standard rate is 0.5, smaller numbers will cause aircraft to
spawn less frequently.  Larger numbers cause aircraft to spawn
more frequently.

Closes #74 and possibly #108

### DIFF
--- a/assets/scripts/input.js
+++ b/assets/scripts/input.js
@@ -330,6 +330,14 @@ function input_run() {
       ui_airport_toggle();
     }
     return true;
+  } else if(prop.input.callsign == "rate") {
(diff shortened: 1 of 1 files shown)


tedrek:
Thanks for the merge.

As a note for anybody coming from the future, #119 changed the standard rate to 1, so for more aircraft use 1.5 for a 50% increase and 2.0 for double the number of aircraft.

PR #121, opened 2015-07-19
Support for lat/long of fixes and runways

tedrek:
While trying to add airports I found most of the information was presented in latitude/longitude coordinates and needed to be converted to the km offset co-ordinates used internally.  This patch changes the airport loading code to support runway and fix locations to be specified in lat/long which are then converted to offset co-ordinates on load.

As an example airport:

```
{
  "name": "Toronto Pearson International Airport",
  ...,
  "position": ["N43d40.63","W79d37.83","569ft"],
  "fixes": {
    "YYZ": ["N43d39m29.00", "W79d37m54.00", "541.0ft"],
    ...
  },
  "runways": [
    {
      "name": ["05", "23"],
      "position": ["N43.685209", "W79.648030"],
      "angle": 57,
      ...
    },
    ...
  ]
}
```

Existing offset coordinates are handled transparently.

Three different coordinate formats are supported: decimal degrees, decimal minutes, and decimal seconds.  All coordinates must start with one of NSEW to differentiate from offset coordinates.  Elevation may be specified in feet or in meters (must be explicit for both) and is converted to feet internally.

### DIFF
--- a/assets/scripts/airport.js
+++ b/assets/scripts/airport.js
@@ -1,10 +1,9 @@
-
-var Runway=Position.extend(function(base) {
+var Runway=Fiber.extend(function(base) {
   return {
     init: function(options) {
       if(!options) options={};
-      base.init.call(this, options);
 
+      this.position     = [0, 0];
       this.name         = [null, null];
       this.name_offset  = [[0, 0], [0, 0]];
       this.length       = 1;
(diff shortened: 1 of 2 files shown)


zlsa:
I honestly hadn't thought of this. I also added support for the `end` key in each runway, to define both ends of the runway. The center position is used, and the length and angle are automatically calculated (but can be overridden with the `length` and `angle` keys).

tedrek:
That's awesome, thanks for merging :)

PR #122, opened 2015-07-20
Make wind at KSFO more realistic

albertyw:
Wind at KSFO pretty consistently comes from the northwest instead of from the southeast.


zlsa:
Thanks! It turns out I just got the angle wrong in the airport file.

albertyw:
Thank you for this repo!  I figured the angle was just flipped by 180 degrees.

PR #127, opened 2015-08-06
Fix collision warning with taxiing aircraft

tedrek:
This fixes my bug where a collision warning is issued between aircraft on the runway and aircraft taxiing from the apron to a runway.


glangford:
+1. Getting "appears on a collision course with another aircraft on the same runway" in a tight loop (possibly the same warning repeated on every position update).

tedrek:
@zlsa, this pull request appears to have been unintentionally closed by #129.  Could you please merge to fix the warning loop?

zlsa:
Sorry and thanks!

Issue #13, opened 2014-12-14
Performance issues


joallard:
I've been playing on my Macbook Air for a few hours, and the fan has been going like crazy since I've opened the game. The battery is taking a big dip either. I'm running Firefox Aurora 36. Is it just me or could this game benefit from a few well-placed perf optimizations?

zlsa:
I've never even attempted to optimize it. I'll work on it.

joallard:
Good job. I've tried profiling a little, but wasn't nearly proficient enough with the architecture to see an obivous fix. [`canvas_update_post()`](https://github.com/zlsa/atc/blob/gh-pages/assets/scripts/canvas.js#L434) came up.

zlsa:
That's where the entire radar view is drawn. There's no view culling yet. The current bottleneck is drawing the text for each airplane; if I have to, I'll move the text drawing to the DOM, but that's a pretty major change.

zlsa:
I've vastly reduced redrawing in version 2.1.4; could you test again?

jvt:
@joallard What year is your MBA? I'm also playing on an Air and didn't notice any performance-related issues in the prior versions.

joallard:
2.1.4: CPU seems to have more breathing space. I think the regular speed has gotten slower though. Might be unrelated.

@jvt Late 2013.

zlsa:
@joallard, aircraft speed shouldn't change; I just redraw the canvas once every 20 frames or so. They might jump visibly when on higher speed settings, though.

joallard:
It may very well just be my perception, caveat emptor, but it feels to me a little more jittery and slowish.

glangford:
A significant portion of the time to draw each aircraft - 60-80% using `console.time()` - is drawing the trailing position dots, not drawing the aircraft itself. This is because `canvas_draw_aircraft()` saves and restores context and recreates a custom `fillStyle` and a filled arc - for every single trailing dot for every visible aircraft. This is very slow. I will send a pull request with an idea for a faster method which is visually similar.

glangford:
The first picture is the existing aircraft trail, the second is a test of a much faster method (rectangles and fixed opacity).
<img width="346" alt="current trails" src="https://cloud.githubusercontent.com/assets/3035114/9170224/0cb7b996-3f2f-11e5-9f48-d2d4dc59debb.png">
[...]

PR #131, opened 2015-08-10
Faster drawing of aircraft tails ; working on #13

glangford:
Simplified drawing of aircraft tails for improved performance/less load on GPU, with similar visual appearance.

### DIFF
--- a/assets/scripts/canvas.js
+++ b/assets/scripts/canvas.js
@@ -262,15 +262,16 @@ function canvas_draw_aircraft(cc, aircraft) {
   if (dpr > 1) 
     trailling_length *= round(dpr);
   cc.restore();
-  for (i = 0; i < aircraft.position_history.length; i++) {
-      cc.save();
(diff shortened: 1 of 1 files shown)


zlsa:
Merged. Would it be even faster to use a path, `rect`, and `fill` separately?

glangford:
> Would it be even faster to use a path, rect, and fill separately?

Not sure. To preserve the current appearance, wouldn't a gradient fill be needed, custom for each trail?

Edit: Maybe don't need a gradient. Perhaps it would be even faster, with something like this:

```
beginPath
loop
  set globalAlpha
  if (i=0) moveTo else lineTo
end loop
ctx.lineWidth = 2
stroke
```

glangford:
So `moveTo, lineTo, stroke` works to draw the trail, but globalAlpha on a point by point basis doesn't preserve the current fading effect -  I just get a solid line. Even a gradient wouldn't give the same effect in all trails because it is coordinate driven, and we want the trail to fade in sequence. 

I'm not an expert on this, please let me know of any other suggestions.

PR #133, opened 2015-08-14
Draw projected aircraft path

### LINKED ISSUE #132: Draw projected aircraft path when selected
One way to help plan aircraft movements is to see where an aircraft will go in future, based on the ATC commands that have been entered.

For example, in following multiple fixes:

waypoints

Or following a heading and landing (the track is brighter when ILS lock is expected to happen):

landing2

These are screenshots from a running prototype. Appreciate any comments or suggestions.

glangford:
Fixes #132

### DIFF
--- a/assets/scripts/canvas.js
+++ b/assets/scripts/canvas.js
@@ -314,7 +314,11 @@ function canvas_draw_aircraft(cc, aircraft) {
     cc.fill();
 
     cc.restore();
-
+  }
+  
+  // Draw the future path
+  if((aircraft.warning || match) && !aircraft.isTaxiing()) {
+    canvas_draw_future_track(cc, aircraft);
   }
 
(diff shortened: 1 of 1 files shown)


zlsa:
I've changed the colors as well; magenta (IMO) stands out too much. I made the colors the same as the airplane (departure/approach), and made the line a bit thicker when ILS is locked.

zlsa:
2015-08-15-08-21-31-am

glangford:
I used magenta since that is the color used in cockpit nav displays, but the colors you substituted definitely look better!

zlsa:
I didn't even know they had tracking (and prediction!) in cockpits yet. Thanks for all the work you've been doing.

glangford:
Glad to help. In the nav displays, I meant for planned trajectory, guided by the Flight Management System. The preplanned track the autopilot will fly is magenta.

PR #136, opened 2015-08-16
Arrival format change and airport JSON cleanup

tedrek:
The major change here is to separate the aircraft heading from the direction in which it spawns for arrivals.  In order to do so the 'heading' key has been re-purposed to apply to the aircraft heading.  The 'radial' key has been added to refer to the direction in which the aircraft is spawned.  If the 'heading' key is not specified it is automatically configured as the opposite direction of the radial giving the current behaviour.  The 'angle' key was removed.

This change is intended to allow a natural format where the spawn location is specified by fix/position instead of direction and the aircraft heading is specified.

There are additional changes which should keep aircraft from spawning on conflicting courses by adjusting altitudes.

Finally there are formatting cleanups which remove DOS line-endings, tab characters, white space at the end of lines and indention variation.


glangford:
In addition to these formatting changes - I am wondering about the spawn frequency settings.

I don't understand how the frequency number in the .json corresponds to the simulation behavior, (it looks like the json numbers get multiplied by 60, and there is also some kind of game frequency setting) but some questions are:
- How have the frequency settings been tuned? When `"frequency": [10, 15]`
  appears in the json, what does that mean in reality?
- What is the maximum total arrival load in an hour for KSFO for example?
- Is there an "end" point for a session, or do planes keep appearing until the user quits?

zlsa:
Frequency is the number of airplanes per minute; if frequency is [3, 3], 3 airplanes will appear per minute (in theory). The game frequency setting is (IIRC) just a speed multiplier.

Planes keep appearing forever.

glangford:
Got it, thanks.

PR #137, opened 2015-08-17
Adjust offsets from lat/long for magnetic north

tedrek:
Using GPS co-ordinates results in angles relative to true north.  Air traffic generally use magnetic headings (ATC vectors, runway angles, etc.), as do all current airports.

To compensate the airport may specify magnetic_north relative to true north in degrees.  Negative numbers indicate an angle west of true north.


glangford:
Nice. Note that for KSFO for example, the magnetic heading for 28L and 28R is 284.1 (the current .json says 283), but with variance of 13.7 degrees east, the true heading is ~ 298. Can you update one of the airport .json in this pull request so there is a complete working example?

tedrek:
Certainly, here's an update to the new KSFO (:+1:) which aligns everything with magnetic north instead of true north.  Specifically note that an aircraft taking off from one of the runways now flies the magnetic heading of the runway instead of a true heading (actually the inverse of the runway heading, but that's another PR :wink:)

glangford:
It looks like the runways have been flipped around as a side effect of this change. If I zoom in to KSFO it looks like this:

ksfo-mag-north

tedrek:
Yes, the runways are rotated 180 degrees when the angle is determined by lat/long.  I'm working on a pull request to fix that issue separately.

tedrek:
#145 fixed the flipping of the runway.  @zlsa I believe this is ready for merge.

PR #144, opened 2015-08-22
More variation to arrival/departure timing

tedrek:
These patches add 2 new scheduling methods for aircraft generation.  A cyclic method which varies the rate over a period, and a wave method which concentrates all the aircraft together as a 7.5 nautical mile in-trail stream of aircraft.

Both methods operate over a period specified in minutes, with an offset in minutes specifying the middle of the peak.  This allows the peaks of various arrival locations to be at different times, and the departures to peak at a different (or the same time) as the overall arrival peak.  When considering offsets between arrivals and departures remember that it takes an aircraft something like 20 minutes to fly from the outer spawn location to the airport.

To use the new types/features a 'type' key is set on the arrival with a value from 'random', 'cyclic', 'wave'.  Existing arrivals/departures without a 'type' key will function as they did before.

The new versions support setting the speed of an aircraft, specifying the range for both radial and heading instead of using a fixed value in the code and most importantly, the interpretation of the frequency key has been changed to units of aircraft/hour instead of minutes/aircraft.

KJFK has been updated to use these new arrivals and departures as an example.  I suspect I've missed a couple of important things so feedback is appreciated.


glangford:
@tedrek - I think there are some great ideas here. I haven't spent a lot of time playing with it yet, but it looks cool. I like the embedded idea in the KJFK example of building the arrivals based on the standard arrival route (STAR) as well.

The down side (if there is one) is that KJFK is problematic in its current form, and there are some unrelated changes we probably want to make so we can take full advantage of these timing variations:
-  KJFK geometry needs to be updated as well (I will do this similar to KSFO if that works for everyone)
- no matter what runway I take off from at JFK, I get a tail wind penalty

If I understand the change to the frequency key - it sounds like the other airport .jsons would eventually have to be updated as well, is that right? At that time, we should figure out what the target frequencies should be based on play testing.

glangford:
This is definitely not the most helpful feedback - but in looking at the approach charts, the Kingston and Lendy STARs are for La Guardia, not KJFK.  :)

tedrek:
Thanks for the feedback @glangford.  I've been building most arrivals based on STARs as that's how it would work in reality at practically all airports which handle this quantity of traffic. The primary goal behind this patch set was to setup a mechanism where the airport can be overloaded significantly at times but there will be a corresponding gap which allows you to catch up so the quantity over time is manageable.  I also wanted to see an things change in such a way that it makes sense to change runway use to a departure heavy setup at times.  Unfortunately I didn't have time to fully play this particular setup of KJFK so I'm not sure how well tuned it is...  Please note you'll need to play for a signifanct period of time to really get the variation which is presented.

The specific issues with KJFK need to be addressed separately, but they do need to be addressed.  I updated KJFK as it's arrivals were completely unusable.

The changes to the frequency key only affect arrivals or departures where there is a type key.  Existing airports don't need to be updated, but it would probably be best to switch everything at some point so there isn't the two drastically different meanings of the key.  I am hoping to do a fairly complete overhaul of arrivals and departures at some point to ensure there is a good balance of difficulties.

glangford:
> The specific issues with KJFK need to be addressed separately, but they do need to be addressed. I updated KJFK as it's arrivals were completely unusable.
> 
> The changes to the frequency key only affect arrivals or departures where there is a type key. Existing airports don't need to be updated, but it would probably be best to switch everything at some point so there isn't the two drastically different meanings of the key. I am hoping to do a fairly complete overhaul of arrivals and departures at some point to ensure there is a good balance of difficulties.

Sounds great - I will get started on KJFK geometry in parallel. I also have a proposal regarding STARs that hopefully will work well with this arrival method. I will document it first as a possible enhancement to get feedback.

tedrek:
KJFK is now playable and demonstrates the variation, it starts off with a heavy departure load while initial inbound aircraft make it to the airport.  After the first five or so arrivals reach the airport area the departures let up and arrivals start arriving from various directions.  The basic pattern repeats on an hourly basis but you can see the variation in about 20 minutes (or 4-5 minutes with warp).

Total counts are for approximately 40 arrivals and 40 departures per hour.

glangford:
@tedrek it looks great!! 

One comment in terms of the target frequency - I haven't tried JFK much in the past so I was definitely not able to keep up to this level of traffic. I would point out though that 40/40 is close to the maximum capacity of the real airport in ideal conditions. So my lack of skill aside, we may want to consider lower numbers given that a single non-expert is doing the task of multiple people.

The JFK airport capacity profile is here:
http://www.faa.gov/airports/planning_capacity/profiles/media/JFK-Airport-Capacity-Profile-2014.pdf

tedrek:
I had made a mistake, the arrivals were about 48/hour.  That has been corrected to 40/hour and departures are approximately 35/hour (unchanged).

I've opened a separate issue to deal with target frequencies for all airports at #156

tedrek:
@zlsa, Thanks for the merge :smile:

PR #155, opened 2015-08-28
Rebalance KSFO traffic

glangford:
This is a rebalancing of arrival and departure traffic for KSFO for review. It targets 36 arrivals and 20 departures per hour (which is not necessarily the right balance). It was play tested using KSFO STARs (manual setting of fixes).

@tedrek - these changes were done in parallel with #144 and don't have the benefit of variable timing. So this can be further improved with variable timing, as well as better adjustment of the frequencies and spawn points. 

The actual airport most commonly sees arrival/departure frequencies of ~ 30/30 according to the FAA. The capacity in ideal conditions is ~ 100 operations/hour.

https://www.faa.gov/airports/planning_capacity/profiles/media/SFO-Airport-Capacity-Profile-2014.pdf

### DIFF
--- a/assets/airports/ksfo.json
+++ b/assets/airports/ksfo.json
@@ -89,20 +89,21 @@
     "airlines": [
       ["ual", 11],
       ["baw", 9],
-      ["cessna", 2]
+      ["ups", 2]
     ],
-    "frequency": [4, 7]
+    "frequency": [2.8, 3.2]
   },
   "arrivals": [
     {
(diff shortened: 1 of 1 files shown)


glangford:
An additional comment on the arrival-departure balance - I set the number of departures lower here since departures are more boring and repetitive from a game play standpoint. But if we are striving for realism, then a balance more like 30/30 is a better choice.

ksfo capacity

PR #157, opened 2015-08-30
Fix bug in magnetic north calculation

glangford:
Fixes are drawn in incorrect locations because of a bug introduced in the magnetic north code.


tedrek:
Oops, thanks for finding the issue.

PR #158, opened 2015-08-31
More realistic turning under ILS capture

glangford:
The current path taken once ILS is locked results in the aircraft being a long way off the centerline even after travelling a long distance. This is an attempt to make the turning more realistic. (@tedrek - would appreciate your opinion on this).

Before:
old ils capture

After:
new ils capture


tedrek:
@glangford, this is a great improvement.  The suggestion I would make for this approach would be replacing the calculation based on model.rate.turn with a fixed value (or in the next line defining the resultant range).  model.rate.turn isn't in use anymore since #118, I really should have removed it completely at that time.

I actually took a look at this the other night, if you look at my branch t-ils-intercept you can see the approach I was looking at.  The drawback is that it's a fairly accurate representation of an ILS intercept, but the game doesn't currently differentiate between visual approaches and ILS (or other) approaches.  My feeling is that until visual approaches are implemented this is too restrictive to be fun.

As reference on ILS this is what I had found:
The primary thing aspect of ILS which is of interest is the instrument in the cockpit only shows +-2.5 degrees lateral from the localizer (the signal is received further out but the needles simply show maximum), therefore heading adjustments will generally only happen within this central cone.  The limitations which ATC operate under are the heading must be within 30 degrees of the localizer heading, altitude must be below the glideslope at intercept and distance is 2 miles outside of the final approach fix by the time the aircraft is on the center line.

glangford:
@tedrek - this is great. My change was just a quick hack to try and force capture to be quicker, but you have a much better fix in the works!

If I play test the t-ils-capture changes, in the scenario I was testing (Modesto Five arrival at SFO with no radar vectors) it won't capture at all. I think this is good though - the aircraft is coming in at too sharp and heading and trying to land on 28R, it probably shouldn't be able to:

t-ils-intercept

For a second test, I give vectors (heading 270) - see here how late it captures (around WETOR):
t-ils-intercept-2

I'm not sure if the capture is too late here, or if it is correct.

glangford:
@tedrek - I think I see the cause of late intercept. From the t-ils-intercept branch:

``` JS
        // lock  ILS if at the right angle and altitude
        if ((abs(this.altitude - glideslope_altitude) < glideslope_window)
            && (abs(offset_angle) < radians(3))
            && (abs(this.heading - angle) < radians(30))
            && (offset[1] < ils)) {
```

Not 100% sure what `offset_angle` is, but is `radians(3)` too stringent here? I don't have a better reference than Wikipedia which describes the localizer coverage like this:

> Coverage: at 25 nmi (46 km) within 10° from CL; at 17 nmi (31 km) within 10° and 35° from CL; at 10 nmi (19 km) outside 35° if coverage is required.
> https://en.wikipedia.org/wiki/Localizer

EDIT: There is a good diagram here:
http://www.pprune.org/tech-log/493857-official-localizer-coverage-25-vs-18-nm.html

glangford:
Also - I don't know how the glide slope window is arrived at in general, but it could use this:

> The centre of the glide slope signal is arranged to define a glide path of approximately 3° above horizontal (ground level). The beam is 1.4° deep (0.7° below the glide-path centre and 0.7° above).

tedrek:
@glangford I'm glad you like what you see in general.

In the first case the relevant restriction is that ATC must give a vector within 30 degrees of the localizer heading for an aircraft to intercept the localizer.  However the aircraft would be capable of intercepting at a higher angle, it's just likely to overshoot in the process.  This should probably be implemented as a negative score for too sharp of intercepts instead of not intercepting at all.

In the second case the intercept is fine, in real life for an ILS approach you need to supply vectors to be established halfway between AXMUL and CEPIN.  The normal procedure as I understand it is to fly the ATC assigned heading until within 2.5 degrees of the center line and then turn onto the center line.

For the late intercept, this is intentional.  offset_angle is the angle between the aircraft's current position and the centerline.  The localizer coverage is significantly wider than 3 degrees and the aircraft can receive the signal outside of the 3 degrees used there.  However, the instruments in the cockpit show a maximum of 2.5 degrees either side of the center line and pilots consequently will fly ATC heading until the instruments show less than maximum deflection. I should have used 2.5 to be completely accurate.

The glideslope window is derived from the glideslope at 3 degrees.  It encompasses the space below the glideslope as well to allow approaching from underneath.

As you found, the approach comes across as quite restrictive compared to what has been in place.  I will try to re-work for a wider capture angle with scoring to discourage high-angle intercepts and we'll see how that plays.  Hopefully I'll have something for you tomorrow.

glangford:
> However, the instruments in the cockpit show a maximum of 2.5 degrees either side of the center line and pilots consequently will fly ATC heading until the instruments show less than maximum deflection. I should have used 2.5 to be completely accurate.

I see your point but I think this 2.5 degree thing is becoming an anachronism. It may be a cockpit limitation of instruments in light aircraft and manual flight, but it is not the case for jet aircraft capturing the localizer with an autopilot engaged. The localizer can be captured well outside of 2.5 degrees. 

From a game standpoint, I agree that this would be too restrictive compared to what is currently in place. But this is a minor point, I will abandon my proposed change!

PR #163, opened 2015-09-06
Runway centerline intercept improvements

tedrek:
This is a follow on from #158.  The aim is to have aircraft intercept and track the extended runway centerline instead of 'funneling' into the threshold. It is modeled after ILS intercept procedures and attempts to balance ILS/aircraft capabilities with ATC restrictions.

The major changes are:
- Reducing capture width from 30 degrees to 10 degrees either side of center.
- Include a negative score with announcement for intercept vectors over 30 degrees from runway angle.
- Set assigned heading to the runway heading on intercept (fixes tracking off the runway on landing)
- Make Aircraft.update() aware of path projection on the canvas to avoid spurious warnings and incorrect score updates.

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -716,16 +716,10 @@ var Aircraft=Fiber.extend(function() {
           this.requested.heading = runway.getAngle(this.requested.runway) + Math.PI;
         }
 
-        if(distance2d(runway.position, this.position) < 10) {
-          console.log("aborted landing");
-          ui_log(true, this.getCallsign() + " aborted landing");
-          prop.game.score.abort.landing += 1;
-        }
-
-        this.requested.navmode = null;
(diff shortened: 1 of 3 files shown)


glangford:
Very nice. 

A minor comment on the topic of the path projection (and the need to add the `projected` parameter to some functions) -  another possibility is to set something like `this.projected = true` rather than passing the parameter around.

This could maybe go in place of the hack I originally put in to not update the strip for projected paths:

`twin.updateStrip = function(){}; // ignore any calls to updateStrip() for the twin`

So `updateStrip()`, `updateTarget()`, etc. could check `this.projected`.

tedrek:
Done, thank you that is a much better solution.

glangford:
Super - thanks for cleaning up the code in `canvas_draw_future_track` as well.

Issue #134, opened 2015-08-15
Defining radar scope vs. control scope


tedrek:
Currently the radar scope is twice the airport ctr_radius or 160 km.  Altitude is ground level to infinity within that distance.  The current control scope is the same distance, with 10,000' being the maximum altitude assignable.

I was thinking of changing the control scope to be the ctr_radius laterally and 0-10,000' vertically.  Aircraft outside of the controlled airspace would not respond to radio commands or show a flight strip, and would be visually displayed in a subdued manner.  All relevant scoring would occur at the control boundary instead of the current lateral 'radar' boundary.  An altitude of 11,000' would be assignable as an implied hand-off to the controller of the airspace above.

I had been thinking hand-offs should be implicit, as they are currently.  Making them explicit doesn't feel fun.

Thoughts? Does this feel right for this game?

glangford:
To make this work I think the spawn rates, positions and tracks for new aircraft would have to be retuned for each airport. For SFO for example, it is currently necessary to issue commands to distant aircraft for things like popping up randomly in the path of a takeoff runway. Also, traffic separation is currently not guaranteed outside of the control boundary, and giving new vectors very early helps a lot. But if there was some greater order imposed (i.e. tighter enroute/area control spawn policy into each airport) then this new approach makes a lot of sense.

tedrek:
I absolutely agree, spawning would need to be tuned.  Likely departure would also need work to keep them from being vectored into an arrival.

Issue #29, opened 2014-12-15
Departing aircraft should have a preferred departure heading


zlsa:
Departing aircraft have a direction they want to depart in; if you're more than 30 degrees off (for example) when they exit the screen, you lose score.

z-adams:
where does it say this heading?

zlsa:
It doesn't, this should be clarified; it's an enhancement.

dominiks:
Would it make more sense to give them a specific heading to leave e.g. 180° and allow some leeway or just specify that they must be south or southeast from the airport when they leave the radar? I'd work in an implementation of such a system.

Also are there altitude requirements when leaving? I remember an ATC game I played on my Palm years ago where each leaving direction had a fixed height attached to it that aircrafts had to have when leaving. No idea of how realistic that is, though.

zlsa:
Yeah, that would be fine. It's just to avoid people like me from having every departure fly north for the simplicity.

daviessm:
atc-sim seems to implement this with aircraft having a required fix to head to after departure. I think that's a fairly sensible way of implementing it; after this point the aircraft could follow its own navigation to its destination - which in our case could be implemented as turning to a heading within 60° of the angle between the airport and the fix and climbing toward the aircraft's maximum altitude.

This could be done with a new command "(r)esume own navigation" with departure score points added for issuing this near the correct fix.

ATC could give the flight new commands afterward if it was heading towards other craft.

It's difficult for this because the sim (@zlsa what is it called? :-) has you perform the tasks of tower, approach and departure (or even more!) which would be different people at a busy airport.

daviessm:
I think this guide from the flightgear flight sim is a pretty good idea of a realish sequence of events:

http://wiki.flightgear.org/ATC_Tutorial#Lining_up_and_taking_off

AlexMcArrow:
I think it would be correct to do the following.
Make a point of several types (waypoint, point of rotation, the point of entry and exit)
- waypoint - just for travel
- point of rotation- just for travel, and, if the list is the latest way the plane, the plane begins to spin around her
- point of entry and exit - the point where the planes arrive and where it should fly for the correct care of the control zone. (in order list, departure planes just have name of point) + degrees heading on enter this point

Also, it will give a very simple way to generate arriving aircraft. And the ability to create a third type of aircraft - transit.

Sorry for my english.

Maverick283:
@zlsa I think this can be closed?

zlsa:
[...]

Issue #147, opened 2015-08-22
Proposal to use standard terminal arrival routes (STARs)


glangford:
This is an enhancement proposal for introducing standard terminal arrival routes (STARs). All feedback is welcome.
### Basic feature

Currently, multiple fixes can be provided in a single command to instruct aircraft to follow a particular path. Often, the same sequence of fixes are typed for aircraft to follow before clearance for landing on an assigned runway.

Rather than typing “ {aircraft id} FIX {fix1} {fix2} {fix3} ...”, common sequences based on a subset of published STARs for each airport can be used.

For example, to mimic the “Modesto Five” arrival at KSFO, we could currently type the sequence of required fixes:

<aircraft id> FIX GROAN CEDES ARCHI MEHTA

Or, we could record this STAR in ksfo.json as an array of fix names, and the user could type:

<aircraft id> STAR MODESTO

(dropping the version number “five” for simplicity). This would result in the same behaviour as navigating the sequence of fixes, and usually reduce the amount of typing. The aircraft navmode would be “fix” when the STAR is assigned, as if all the fix names were entered.

Entering the STAR will not always be shorter to type; for example, the Golden Gate Six arrival consists of just the fixes PYE and DUXBY. A more compact STAR name could be used (eg. “GGATE”).
### Additional supporting features
- KSFO can be used as a test airport as it already has the required fixes; additional fix points can be added for other airports, as required
- supported STARs for an airport could be drawn on the canvas as thin solid or dashed lines connecting the fixes, so there is a visual reminder
- the flight strip could reflect the current STAR, as well as the current fix. In the above example, “MODESTO asterisk GROAN” could be displayed in the strip after the STAR command is entered (rather than just “GROAN”. Later, “MODESTO asterisk CEDES”, etc.
- for spacing traffic in the real world, shortcuts are often provided which deviate from the published routes. An additional command called “DIRECT” could be added to provide a shortcut, while maintaining the list of fixes. For example, “DIRECT ARCHI” would delete any intervening fixes, but retain ARCHI and MEHTA on the Modesto Five STAR. This is the same command ATC would give.
- to get a reminder at selected fixes, a controller can ask for an aircraft to report at selected waypoints/fixes. An additional command called “REPORT” could be added. For example, “REPORT ARCHI” would cause a message to be displayed when that fix is reached.

akymakr:
Great enhancement.
Few suggestions listed here: https://github.com/zlsa/atc/issues/128

Keep on good work!

[...]

PR #164, opened 2015-09-07
Separate control radius from radar range and add departure destinations

tedrek:
This is a pull request for feedback.  There are further changes required before merge.

Two major changes:
- Aircraft outside of airport.ctr_radius are not controllable - #134
- Aircraft have departure destinations - #29

I've tested this on KDBG only, all other airports will require some tuning before being playable.

The first change uses the center radius as the control zone, aircraft outside of the terminal area do not respond to commands and are displayed in a subdued manner.  Scoring for arrivals/departures occurs at the area boundary instead of the radar range.  To support this aircraft fly directly away from the airport upon leaving the area boundary.
One 'feature' which needs feedback here is the ability to order an aircraft to climb above the area ceiling to depart.  To use this clear an aircraft to 11,000 feet.

The second major change is adding aircraft departures, by default everything will depart due north.  The airport.json can specify different headings for departures.  A blue arc 10 degrees wide is shown at the area boundary to show where the aircraft must fly to score a valid departure.
Departure directions are included in this PR as a method for separating departures from arrivals outside of the terminal area.

A logical further development will be to eventually extend departure specification to be like arrivals.  This would allow some airlines do depart in certain directions more than others for example.

Before merge, all airports need to be updated with departure directions and arrivals to be certain they won't conflict with either departures or other arrivals prior to reaching the terminal area.


glangford:
Some initial comments:

#### General
- shifts the balance of play between departures and arrivals
- allows greater focus on the control area
- future - do we want to add the ability to schedule heading or altitude changes to be followed only after a fix is reached?

#### Departures
- both more interesting from a game play standpoint and more realistic
- the blue destination arc is a good visual reminder - not visible when zoomed though
- should the arc be bigger?
- scoring could be less punitive; getting a departure safely off the ground but missing the arc is still a good thing
- no opinion on the 11,000 feet clearance at this point, but I am guessing that in some instances departing aircraft may be initially cleared to a low altitude under arriving aircraft, then later assigned a fix and higher altitude to leave

#### Arrivals
- more realistic to only be able to control aircraft within a defined radius
- less zooming in and out required to play since beyond range aircraft can't be controlled
- spawn locations, headings and spacing become more critical since player has no recourse to adjust them beyond range
- perhaps arriving aircraft could be assigned navigation fixes at spawn time

tedrek:
A future ability to schedule commands after or at a fix would likely be desirable.  I think #147 could naturally create such a capability, it would also be the place to assign fixes at spawn time.

For departures I would like most airports to eventually have actual fixes instead of just a heading/radial.  This would make the arc visibility and size much less important.  The actual implementation would be nearly identical, but instead of a radial shown on the flight strip it would be a fix.  The fixes would be tuned to be very near the terminal boundary.

Scoring for failed departures should probably be reduced, it is equivalent to a failed arrival, which doesn't take into account missing by a small margin.

I will review and update all arrivals for all airports so there aren't any conflicts outside of the terminal area.  The assumption required of course will be that departures generally hit their destination radial.

glangford:
> For departures I would like most airports to eventually have actual fixes instead of just a heading/radial. This would make the arc visibility and size much less important. The actual implementation would be nearly identical, but instead of a radial shown on the flight strip it would be a fix. The fixes would be tuned to be very near the terminal boundary.

Makes sense - I am just finishing an update to KMSP and I will add departure SID fixes to that update. (KJFK and KSFO should be updated with additional fixes also).

tedrek:
All airports now have departure destinations, I didn't do any research I just paired a destination with each arrival.  Two new airlines were added, lightga and fastga.  Airports have been updated so arrivals are separated between jet aircraft and the slow prop aircraft of lightga.  Lightga arrivals should either be infrequent or have a speed set of 80-120 to keep aircraft from overtaking the previous one.  The only reason there should be conflicts outside the terminal area is if a departure is vectored outside of it's destination.

Departure failures score a -2 instead of a -20.

nielsvdweide:
I still can't see these departure destinations. Did you implement this already?

glangford:
It's in tedrek's development branch t-departure-destinations - it isn't merged yet.

nielsvdweide:
I am terribly sorry, I still don't know how this thing works! Thanks for pointing this out to me!

glangford:
> I am terribly sorry, I still don't know how this thing works! Thanks for pointing this out to me!

No problem, we are all learning from each other. :-)

tedrek:
@zlsa, this is ready for merge from my perspective.

zlsa:
I've been very much behind lately - thanks for pinging me @tedrek, I wouldn't have noticed otherwise.

tedrek:
no worries @zlsa, thanks for making it in the first place and continuing to be the steward :smile:

PR #165, opened 2015-09-07
Show degree directions for the selected aircraft

ivant:
This also requires minor fixes to:
- input, to prevent unintended persistence of the directions once an
  aircraft is deselected by pressing Esc.
- canvas, to improve the resolution of the directions.

### DIFF
--- a/assets/scripts/canvas.js
+++ b/assets/scripts/canvas.js
@@ -24,9 +24,6 @@ function canvas_init_pre() {
 function canvas_init() {
   "use strict";
   canvas_add("navaids");
-  canvas_add("info");
-  canvas_add("aircraft");
-  canvas_add("compass");
 }
 
 function canvas_adjust_hidpi() {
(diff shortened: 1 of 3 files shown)


ivant:
A screenshot of the directions in action (the layer is only active when an airplane is selected): [here](https://goo.gl/photos/ShwKbdighsN2NtuP6).

tedrek:
This is an interesting idea.

I generally play zoomed out to see the whole center and don't pan or zoom much.  Therefore commands are often given to aircraft near the edge of the screen.  In this modification it results in an unreadable bunching of numbers/marks on the edge nearest the aircraft.

glangford:
I think this idea could be useful to newer users.

I try to assign fixes as much as possible, and after headings are assigned over time I think there is a training effect. In the early days, the brain is trying to figure out what number matches the desired direction, but over time it becomes automatic.

Side comment: I liked the use of a separate canvas, and it looks like the existing code was set up with multiple canvases but they aren't used.  The monolithic canvas.js could be partitioned so that features like this could be added easier, and redrawing would be minimized. For example http://stackoverflow.com/questions/3008635/html5-canvas-element-multiple-layers.

ivant:
I agree about bunching when the aircraft is near the edge of the screen, however this direction is usually the opposite one from the direction that you're interested in, so it doesn't matter that much. In terms of polish, I considered simply avoid drawing the numbers (and possibly the degree marks) if they are too close to the aircraft, however I decided to do it as a separate commit.

Fixes work great for the airports that have enough of them, unfortunately some airports have rather unhelpful fixes (EDDH is an example). Also it's occasionally useful to be able to turn the aircrafts ever so slightly when they are in danger of passing too close shortly.

ivant:
I'm on the fence about using separate canvases (with the current way `canvas.js` is set up to do things). Unfortunately I miss the context for why they are present in the code but not used.

One possible reason of why multiple canvases could be bad could be the increased amount of memory/CPU needed to handle them, but I don't know how serious that problem is.

Partitioning different things into separate canvases could be a useful change in terms of extensibility, however it seems that most of the canvases would need to be updated on every frame.

zlsa:
It was originally intended to speed things up so that the runway, etc. didn't have to be continuously redrawn, but since panning and zooming have been added (and the runways are very fast), it turned out to be slower (on my computer) to use multiple canvases.

ivant:
OK, sounds like I should rework this change to just use a single canvas.

tedrek:
I did an experiment with multiple canvases at one point and found it was slower on my system as well.

erikquinn:
As of commit 0717412, this _has been_ reworked to use a single canvas as requested. This is a helpful feature for beginners who have trouble choosing good headings, and though it will soon be made a toggleable option (disabled by default), I see no reason not to merge these changes. Performance with and without this change is showing no appreciable difference.

Issue #68, opened 2014-12-21
Changeable weather


daviessm:
Winds could change direction forcing controllers to route planes to a different runway. High crosswinds could mean (some) planes could not land.

Poor visibility could result in VFR flights being unable to land.

sixeyeco:
@daviessm: variable wind direction would be amazing. diverting a plane that is not able to land due to excessive CW to another airport should add points if ordered in a timely manner (e.g. prior to hitting localizer). rain could increase landing time/extend landing distance forcing adaptive taxiing practices. thunderstorms could drift through radar necessitating diverting aircraft to avoid weather.

erikquinn:
Piggybacking on this one (just the wind part), I had the following to say (moved from closed issue #325): 

---

It would add some realism if the wind could vary a bit. Not in the sense of necessitating a runway change necessarily, just that it has shifts and swells of direction and speed. I'm not sure how this is best implemented... what about based on a few levels of Math.random() values, like this:
- Net change from 0h0m->1h0m:   _[(Math.random() = 0.8513) x 2 - 1 = **+0.7026**] x 18kts = **+13kt**_
- Net change from 0h0m->0h30m:   _[(Math.random() = 0.6912) x 2 - 1 = **+0.3824**] x 9kts = **+3kt**_
- Net change from 0h0m->0h10m:   _[(Math.random() = 0.2860) x 2 - 1 = **-0.4280**] x 3kts = **-1kt**_
- Wind @ (t): uhhhhhhhhhhhhhhhh???? 

This way, the changes would be random, but with a clear overarching trend. Somebody more well-versed may have a more intelligent solution, utilizing some of the functions from util.js that I still can't quite figure out what they do. If so, bring on the suggestions!

Maverick283:
Picking up @harp71 comment from the other issue:

```
"wind": {
  "angle": [[250, 280], [70, 105]],
  "speed": [5, 20] 
},
```

would be a great solution I think. Now all there is left to write is a algorithm as suggested by @erikquinn, that can make this wind "come alive". Though I think that a very rare chance of completely off winds would be great and realistic as well.

Maybe this [MATLAB Thread](http://de.mathworks.com/matlabcentral/newsreader/view_thread/318455) may help.

erikquinn:
Also, as mentioned in #630, it might be interesting to have the option of updating the game with the actual live wind conditions.

erikquinn:
[...]

Issue #156, opened 2015-08-28
Balancing arrival/departure rates for realism and fun


tedrek:
The current arrival and departure rates seem to be inconsistent between different airports.  This issue is trying to develop a set of guidelines for arrival and departure rates in the future, and a proposal to rebalance arrival and departure rates of existing airports to meet the needs of the community.

Currently difficulty scales with the aircraft rates, these have been historically chosen by which airport is played and more recently can be adjusted with the rate command.  Each airport is assigned a difficulty level, however that is not exposed to the player at this time.  Following is a table of the current airports with their operations in aircraft/hour.

| Airport | Difficulty | Arrivals | Departures | Contributor |
| --- | --- | --- | --- | --- |
| kjfk | hard | 12 | 30 | psycho202 |
| sbgl | hard | 13 | 7 | MaicomMR |
| ebbr | hard | 25 | 30 | stijnsanders |
| eddh | medium | 25 | 30 | larsen0815 |
| ksfo | hard | 36 | 20 | zlsa & glangford |
| eham | hard | 36 | 40 | Sander Ufkes |
| sbgr | hard | 39 | 40 | MaicomMR |
| kdbg | hard | 49 | 30 | zlsa |
| uudd | hard | 51 | 20 | MaicomMR |
| kmsp | hard | 55 | 30 | Joe Turgeon |
| ksra | medium | 60 | 30 | zlsa |

There are a couple things of note:
- ebbr and eddh have identical arrivals (copy and paste in the .json) yet differing difficulties.
- kjfk is not complete and should be disregarded
- kdbg and kmsp have identical arrivals with a couple of frequency changes.

From the above it appears the playable range in terms of operations is considered to be 20/hour to 90/hour.  Arrivals are preferred over departures.  The number of operations isn't as significant as the airport to some players as evidenced by new airports with identical arrivals to existing airports.

The proposal which I would start off with would be:
- Model airports based on real-life 80% capacity (approximately)
- Rate airports on a scale of 1-5 for difficulty:
  *\* 1 = 0-20 arrivals
  *\* 2 = 20-35 arrivals
  *\* 3 = 36-50 arrivals
  *\* 4 = 51-70 arrivals
  *\* 5 = 71+ arrivals
- Expose the difficulty rating beside each airport as 1-5 aircraft icons (✈)
- Prefer departures as equal or less than arrivals
- Keep the rate command to allow players to tune a given airport to their preference (hopefully expose it through the GUI at some point)

Thoughts?

[...]

Issue #128, opened 2015-08-07
Far from Tower Controller


akymakr:
**We are just a Tower Controller, right?**
I have to clear an attitude, move the aircraft from the ground and the sky, and also clear some aircraft to land or takeoff? If this is a small airport ... OK but ... this is JFK, SFO and MSP! "Hire" someone to handle those OK? We just wanted to clear somebody to takeoff or land. Or we just want to line up some aircraft and then pass it to the Tower.

**Too many aircraft and they come from anywhere and low**
If I am the tower controller, I am gonna kill the guy who is handling the approach. You just can't let an aircraft coming from anywhere, without lining them up and pass it to the Tower in 50 miles away from the airport (also we can't tell them to "contact departure 125.7" or something). Can you just lower them down and create some fix that the aircraft should be born at?

**Command suggestions**
Go around, they don't understand. (abort)
Hold at XXXXX (waypoint) at YYY radial, 10 mile legs, left hand turns, they don't understand.

**Other suggestions and issue**
Well, the waypoint is lacking of many airports, not enough to provide SID or STAR. Also the coordinates is not accurate. We can't "Cross DUMBA at 3000 or above, cleared ILS 28R approach". DUMBA is just too far away from the airport.

Eh ... provide SID and STAR (not so urgent)?
Also a controller selection for DC, GND, TWR, DEP, APP? (I think not only me want this one)

Some callsign used normal alphabet but not the NATO one.

I could help to handle those JSON if you don't mind :)

harp71:
Second that ...

And there has to be a "mission, target, goal" (call it what you want) for departing flights

glangford:
@akymakr @harp71 try the new version of KSFO and post any comments. Aircraft now arrive with fixes set; see ksfo.json. 

Departures now have a goal (all airports).

harp71:
Oh, the departure goal is represented by the blue marker on the outer circle? That´´s better :)
I´m used to (from other ATC sims) having a departure goal before taking of, so you can choose the runway accordingly. And the goal is usually not a area but a fix-point.
But its fine that way for me :)
Finally there is sense to departures.
Now we need SIDs :))

Maverick283:
@harp71 SIDs have already been mentioned earlier, yet STARs seemed to be more important for the moment. The guys programming all this can always do only so much ;)

harp71:
i´m gratefull no doubt. I wish i could contribute more than suggestions ...

[...]

PR #169, opened 2015-09-07
Implement variable wind at KSFO

tedrek:
Here's an approach for fixing #68.  The basic principal is to build a wind rose for the given airport, I used data from http://www.enviroware.com/METAR/METAR_WindRoses_2014_maps.html for KSFO.

A wind direction and speed are selected with with weightings, then over a 3 hour period the wind shifts to a new direction and speed, repeat.

I increased the crosswind limit from 10 to 15 for normal and 20 to 25 for major.  (Of course this can't generate cross winds over 24 knots so I'll need to fix that, I had planned to make the top tier 20-40 knots but forgot.)

Any feedback appreciated.


zlsa:
What do you think about making the delay more like 30 minutes or so? I honestly don't see anybody playing the game for three hours and watching the wind direction move.

tedrek:
I initially had a 60 minute delay, that seemed too fast to be realistic.  It also made the wind switch fast enough that by the time an aircraft had made it's way from the center boundary to a runway that runway could be already unusable.

Looking at it now, I think there should be two different variation methods, one which slowly varies the speed and direction and one which causes quick and drastic changes.  This is something which would probably benefit from some sort of configuration where the player can choose whether a drastic wind change is common (50% of games or better) or rare (5% of games or less) for example.

glangford:
My two cents - I think this is a terrific idea in general, but I worry a bit about timing and introducing a feature like this before the baseline is more stable. With #156, #134 and #128 (and #147 should we decide to do it) does it make sense to try and reach a stable place in terms of playable arrival/departure rates first?

tedrek:
@glangford, that's a very good point about timing.  I'll leave this for now and come back to it later.

glangford:
I am trying the enviroware link above, and getting very slow response. Here is another data candidate:

http://windhistory.com/station.html?KSFO

erikquinn:

Closing this PR.

Issue #170, opened 2015-09-08
Collision on landing fills screen with messages, deducts ~5000 pts


joallard:
I think I set two planes to land face to face on the same airstrip. Needless to say, -5000pts is a little heavy for a penalty.

tedrek:
I tried to duplicate the bug but couldn't get two aircraft to collide on landing, I did manage to get two aircraft to hit each other more than once though.

Looks like there is an issue with hits not changing the aircraft state properly, perhaps to do with runway tracking.  I'll try to look at it in the next couple of days if somebody else doesn't get to it.

glangford:
I also tried to duplicate the bug but was unsuccessful.

> Looks like there is an issue with hits not changing the aircraft state properly...

I looked through the code with that thought in mind. Here is a pathological case to consider; _3_ aircraft in close proximity, _all below 1000'_. Call them `this`, `other1` and `other2`.
- In `updateWarning()`, `this` and `other1` are landing face to face on the same airstrip and a warning is generated. `this.warning` is now true.
- Going through the `for` loop a second time, `this` and `other2` are close, but not satisfying a warning. `this.warning` now is set to false; see the annotated code below.
- the next time an update occurs, `this.warning` is reset to true for `this` and `other1`, and the negative score is repeated

``` js
          // Check for the same runway, headings differing by more
          // than 30 degrees and under about 6 miles
          if ((this.requested.runway != null) &&
              (airport.getRunway(other.requested.runway) ===
               airport.getRunway(this.requested.runway)) &&
              (abs(angle_offset(this.heading, other.heading)) > 0.5236) &&
              (distance < 10))
          {
            if (!this.warning) {
              ui_log(true, this.getCallsign()
                     + " appears on a collision course with " + other.getCallsign()
                     + " on the same runway");
              prop.game.score.warning += 1;
              this.warning = true;
            }
            warning = true;
          } else {
            this.warning = false;  // <---- warning is cleared  ******
          }
```

Not sure this is the cause, but it could be one path through the code causing the warning state to flip flop, and trigger cascading warnings.

EDIT: Here is a link to the code of interest
https://github.com/zlsa/atc/blob/gh-pages/assets/scripts/aircraft.js#L1178

tedrek:
Nice work, so here is a test case which shows the issue:
- Line up two aircraft on parallel runways for takeoff (`xx1 w 36l c 1`, `xx2 w 36r c 1`)
[...]

PR #173, opened 2015-09-10
Correct assignment of aircraft.warning

tedrek:
The quick fix to #170 and how it should have been done initially (by me).


glangford:
+1

Issue #175, opened 2015-09-14
Select all airplanes.


nielsvdweide:
I would like to implement a way to select all airplanes. This way you can see what every airplane is doing and if they are going to cross each other. I came across this when i type in a letter some airplanes will be selected including a picture of where they are heading and going. I would like to add a command which shows the path of all airplanes! We could use the word "all" to highlight all airplanes.

<img width="800" alt="schermafbeelding 2015-09-10 om 14 09 53" src="https://cloud.githubusercontent.com/assets/14002770/9845616/9894e11a-5acc-11e5-9fda-669e39320e61.png">

As you can see, 3 airplanes are selected, but i would like to see a new function where you can select all airplanes!

I would like to have some thoughts about this! Let me know

Maverick283:
As STARs and thus routes are about to be implemented, a line following the whole path of the route each airplane takes would be a great help in figuring out where each aircraft is going, as STARs are assigned by the system rather then the user.

nielsvdweide:
Any programmers that can implement this?

<img width="1911" alt="schermafbeelding 2015-09-16 om 15 23 10" src="https://cloud.githubusercontent.com/assets/14002770/9905843/a521b022-5c87-11e5-83e9-a1bfa375841a.png">

As you can see now all airplanes have been selected :)

glangford:
> Any programmers that can implement this?

@nielsvdweide - I like the idea, I hope to have some time to look into it.

glangford:
How do you feel about a small change to this feature? Instead of the word "all" as a distinct command, how about typing asterisk ("*"), without having to press return/enter? This is very easy to implement, and selects all airplanes with less typing.

glangford:
> As STARs and thus routes are about to be implemented, a line following the whole path of the route each airplane takes would be a great help...

@Maverick283 - good idea. Note that this could be useful even before STARs are implemented, since a series of fixes can be assigned today by providing multiple fixes in one FIX command. Can you open a separate issue to track this?

nielsvdweide:
@glangford You could use any command! It doesn't really matter what it is! Typing "/ " for instance does the same and is easy to reach on the keyboard too! I think that it's important to pick a command which is easy accessible on the keyboard. 

What do you mean with: "Note that this could be useful even before STARs are implemented, since a series of fixes can be assigned today by providing multiple fixes in one FIX command. Can you open a separate issue to track this?"

glangford:
[...]

PR #176, opened 2015-09-16
Implement select all aircraft, fixes #175

glangford:
All aircraft can be selected (and their future tracks can be viewed) by entering '*'.

PR #177, opened 2015-09-17
Extend future track line with scheduled fixes

glangford:
Draw a dashed line representing any future fixes which are scheduled after 60 game ticks. If multiple aircraft are selected, their future tracks are composited in a (hopefully) meaningful way. Implements an idea suggested in #175.


glangford:
Here is an example screen shot of the feature when combined with #176.

extended track

Issue #174, opened 2015-09-10
Multiple Suggestions: 1) Holding 2) Routes between fixes 3) Select all airplanes


nielsvdweide:
Hi there! I have got a couple of suggestions to make this atc simulator a bit more real so let me know what you think.

1) Holding patterns. 
At this point, when you send an aircraft into a hold it will just circle either left or right. In commercial aviation a hold consists of a 180 degree turn, then 1 minute outbound, then again a 180 degree turn and another 1 minute leg. They make this 1 minute straight flying so the aircraft equipment will not suffer from conditions caused by being in a constant turn. Therefore, i would like to suggest to make the holding pattern look more real.

<img width="200" alt="schermafbeelding 2015-09-10 om 14 08 46" src="https://cloud.githubusercontent.com/assets/14002770/9787976/33e9aa5a-57c7-11e5-8a98-249991ce2ce1.png"> 

![holding_pattern-e1400680150992](https://cloud.githubusercontent.com/assets/14002770/9788008/70625d4c-57c7-11e5-81e1-c0ae2b51c1fe.jpg)

Eventually, we can also implement this holding technique to make aircraft hold over a fix instead of just circling in relation to nothing. For example: ACA627 f RTM then hold right. The aircraft will fly to rotterdam and then start a holding pattern over that fix. Ofcourse this is for later concern but then we also need to implement the correct entry of a holding. More later! 

2) Routes between fixes
Another thing i came across is when you make an airplane fly to a couple of fixes, including some of the with a big angle to turn. Normally, an airplane will fly as effective as possible, so they will turn a bit early of the fix to get on a straight line between the two fixes. 

<img width="500" alt="schermafbeelding 2015-09-10 om 14 08 29" src="https://cloud.githubusercontent.com/assets/14002770/9788101/12d0d144-57c8-11e5-83b8-04f7bd8ad272.png">

As you can see i made the airplane fly to RTM and then to STD. I would like the airplane to fly a straight line between the to fixes, yet the airplane overfly's the fix and then starts the turn, which causes the airplane being off-track. This only happens of course when the angle to turn is big enough to cause this problem. The way to fix this is to make a calculation based on airspeed (as turn radius is a function of airspeed) and angle the airplane has to turn, so the airplane comes on a straight line between the two fixes. I don't know if this is hard to fix, it is just my suggestion! 

3) Selecting all airplanes
[...]

PR #183, opened 2015-09-22
Implement fly-by fixes and turn anticipation

glangford:
When navigating from fix to fix, the current physics rules fly the aircraft within 0.5 km of one fix before turning to the next. When larger direction changes are required, aircraft will often anticipate the turn and “fly by” the waypoint rather than flying close to the fix, particularly when larger changes of direction are required. This avoids overshoot of the fix. See #174, /cc @nielsvdweide.

For aircraft navigating between two fixes, this change calculates the turn initiation distance as an aircraft FMC would, and starts the turn to the next fix at that point (to a maximum distance of 10km from the fix). 

For course changes of < 20 degrees at 200 knots, there is no noticeable change since the turn initiation distance is small. At 90 degrees/200 knots, the turn starts 2.4km from the fix and increases from there for sharper turns.

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -1048,7 +1048,9 @@ var Aircraft=Fiber.extend(function() {
         var fix = airport_get().getFix(this.requested.fix[0]);
         var a = this.position[0] - fix[0];
         var b = this.position[1] - fix[1];
-        if(distance2d(this.position, fix) < 0.5) {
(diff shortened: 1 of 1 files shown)


zlsa:
Nice work!

MaicomMR:
:+1:  Wow man, very very nice work!!!!

glangford:
@zlsa, @MaicomMR Thank you!

Issue #185, opened 2015-09-23
"Leaving radar coverage" bug when navigating to a fix


glangford:
The message "leaving radar coverage outside departure window" was given for UPS 624 pictured below; it is supposed to fly to LOCKE as the dashed line indicates, but it suddenly changed direction (uncommanded) as the message was generated.

This is likely triggered by UPS 624 reaching 10,000 feet.

It looks like a bug in aircraft.js `crossBoundary`; departing aircraft are automatically switched to "heading" mode.

![ups 624](https://cloud.githubusercontent.com/assets/3035114/10049607/3c324b12-61e6-11e5-8ba1-dc78a5a95b3f.png)

tedrek:
That is expected and intentional behaviour for the code as written.  I'm not saying it's the correct behaviour for the game.

The intent with the radar coverage change was to reduce the controlled area to a defined cylindrical volume of air space with a ceiling.  The relevant change here is that you can clear an aircraft to 11,000 as an implicit handoff to the sector above you.  Once an aircraft leaves your sector the next sector vectors it directly away from the airport and climbs to 20,000 (at which point it should generally be removed from the simulation).  This is realistic in that the next sector will issue commands relevant to it's situation regardless of what you had issued.  The only way I saw to get around this is to have an explicit hand-off command which you would have to enter for every departure at the appropriate time, that doesn't sound like fun to me.

I have ran aircraft at 10,000 without them leaving control, so I suspect you cleared UPS624 to 11,000 or higher (not necessarily intentionally).  My best suggestion as a fix would be to disallow climbs out of the airspace, reverting to the original vertical behaviour.  When the option dialog gets implemented, then have a switch which allows climbing out of the airspace to leave the sector.

The reason for switching to a heading mode after leaving coverage was to avoid conflicts with arrivals.  By flying a radial directly away it guarantees lateral separation from arrivals if it leaves within the departure arc as all arrivals and departures are configured to be separate.  It also is a guaranteed mode to remove the aircraft from the game once it is out of the players control, leaving in fix or hold mode could result in an aircraft circling indefinitely.

glangford:
> My best suggestion as a fix would be to disallow climbs out of the airspace, reverting to the original vertical behaviour.

Alternately - it sounds like we can just clear them to 10,000 until they leave the horizontal airspace, and this issue wouldn't occur?

tedrek:
That is correct, clearing to only 10,000 will keep them within the controlled area.

[...]

PR #187, opened 2015-09-23
Add ability to spawn arrival aircraft with STAR fix list

glangford:
Proposed set of changes using KSFO as a test airport (checked items are done):
- [x] Add ability to spawn arrival aircraft with a list of fixes, rather than a heading
- [x] Update KSFO to use STARs
- [x] Test KSFO
- [x] Update KSFO with cyclic/wave arrivals (#144)
- [x] Adjust KSFO departure and arrival rates (#156)
- [x] Play test
- [x] *Add error checking for invalid fix names
- [x] *Ensure aircraft spawn outside of ctr_radius

Suggestions welcome. (I would ordinarily do this as a set of commits, then rebase and squash into a single commit and send a pull request at the end. But I wanted to try it in stages.)

### DIFF
--- a/assets/scripts/airport.js
+++ b/assets/scripts/airport.js
@@ -209,7 +209,7 @@ zlsa.atc.ArrivalBase = Fiber.extend(function(base) {
       if(start_flag) // At start, spawn aircraft closer but outside ctr_radius
         distance = this.airport.ctr_radius + random(10, 20);
       else
-        distance = 2*this.airport.ctr_radius - random(2, 18);
+        distance = 2*this.airport.ctr_radius - 2;
       var position = [sin(radial) * distance, cos(radial) * distance];
 
       var altitude = random(this.altitude[0] / 1000,
--- a/assets/scripts/airport.js
+++ b/assets/scripts/airport.js
@@ -98,7 +98,7 @@ zlsa.atc.ArrivalDefault = Fiber.extend(function(base) {
         airline:   choose_weight(this.airlines),
         altitude:  altitude,
         heading:   this.heading,
-        fixes:     this.fixes.slice(),
+        fixes:     this.fixes,
         message:   message,
         position:  position,
         speed:     this.speed
--- a/assets/scripts/airport.js
+++ b/assets/scripts/airport.js
@@ -128,6 +128,7 @@ zlsa.atc.ArrivalBase = Fiber.extend(function(base) {
       this.altitude = [1000, 1000];
       this.frequency = [0, 0];
       this.heading = null;
+      this.fixes = [];
       this.radial = [0, 0];
       this.speed = 200;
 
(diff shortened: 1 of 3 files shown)


assets/scripts/airport.js
```diff
@@ -95,6 +98,7 @@ zlsa.atc.ArrivalDefault = Fiber.extend(function(base) {
         airline:   choose_weight(this.airlines),
         altitude:  altitude,
         heading:   this.heading,
+        fixes:     this.fixes.slice(),
```
tedrek: Should the aircraft do the clone instead of the caller?  I generally think of arguments as being immutable unless it's explicitly passed as a mutable value.

tedrek:
Very nice :+1:

assets/scripts/airport.js
glangford: A good point - not knowing the whole code base, I guess I was being slightly paranoid not knowing what other code might do with the options. The defensive approach was to clone it and prevent any means of it being modified.

glangford:
@tedrek What do you think of the following approach to partly address #185:
- Rather than blindly using the vanilla STAR as specified in the airport .json, at spawn time the distances of fixes from the origin are checked. Any fix which is further than the aircraft spawn point is eliminated from the aircraft fix list. If all fixes are eliminated, the aircraft changes to heading mode.

tedrek:
That sounds simple and functional.  While you are doing that, you should run across the code which sets the distance using offset and can ensure the minimum spawn distance is outside the ctr_radius.

As an optimization, that check can be done only in the initial load branch of spawning.  The only place where it will look funny is for a STAR with legs which are more lateral than radial, I don't think those are going to be very common and a little variety is good anyway :wink:

glangford:
After some testing, KSFO is now at 38 departures and 36 arrivals / hour. The max for the real airport is 45/55 in good conditions. Although 36 seems low as a percentage, it is busy for one person at peak.

glangford:
If the team is ok with the revised code for choosing the spawn distance, I think this is now ready for merge.

Maverick283:
Will this change the complete arrival spawns? I noticed some planes flying by the east side of the airport with a heading of about 200°. If aircrafts still spawn with headings, this might be something to fix before merge.

Maverick283:
> 98 -      "radial":    45,
> 99 -      "heading":   200,

I guess it was them two lines, and as far as I can see these are deleted, and no aircraft will spawn there anymore, not sure if that is correct (as that is basically all the code I have looked at and I am so not familiar with the rest of it)?

glangford:
@Maverick283 - for KSFO, all arrivals spawn with fixes. For the RISTI 4 arrival, you will see in the airport .json that a heading and a fix list are assigned; this is so that they spawn pointed south, and don't turn into the coverage area before leaving again to get to MOVDD.

glangford:
> I guess it was them two lines, and as far as I can see these are deleted, and no aircraft will spawn there anymore, not sure if that is correct?

The aircraft formerly spawning at radial 45 are now at radial 60. This is intentional.

nielsvdweide:
Is this still being tested?

glangford:
> Is this still being tested?

It is ready for merge as far as I am concerned. @zlsa, @tedrek - if you have suggestions for changes, or any concerns, let me know.

zlsa:
Looks good, nice work! Merging.

glangford:
Good stuff, thanks for the merge! Appreciate any additional comments when others have a chance to try it - note that it is KSFO only for now. (Bonus - you can use fixes for departures too, some new fixes used in SID routes were added).

Issue #28, opened 2014-12-15
Move aircraft tags


zlsa:

sixeyeco:
I'm fairly certain you have all the input you need on this but...when I vector extremely tightly the tags can be misleading. since the (craft's) dot gets big and the tag is highlighted then the tag itself could be much further away. 

![image](https://cloud.githubusercontent.com/assets/179496/6193499/a30a0746-b370-11e4-89d6-31e78d325afb.png)

when craft lock on to ILS I'd prefer their tags be a lot further away. craft should already 6-7k @ 200 knots so what i concentrate on are the trailing dots (is that also its wake?). 

also current placement appears to lead the craft, wouldn't it be better be to somewhat trail the craft? if its trailing dots, another craft, or another tag comes closer than, say 20 pixels, then it moves accordingly (e.g. move 5 pixels away from obstruction).

PR #194, opened 2015-09-27
Datablock in DOM and movable

tedrek:
Change the aircraft info to be a DOM element.  Also change it to be draggable fixing #28.

The current graphics are... ugly.  @xenocide0, is there any chance you could create .png versions of the original ILS lock images with transparent backgrounds?  I did a quick copy paste from a screen grab but getting rid of the background color was beyond my abilities at the time.

The performance seems to suffer noticeably.  @glangford, could you take a look and at least confirm it's an issue on your machine (if not give it a profile and a pointer).  My suspicion is the number of DOM elements which need to be moved and the resulting compositing are the bottle-neck on my machine.

I went to the DOM to allow an easy way of capturing the click and drag for tags as opposed to the pan function.  With current performance though, it may require a re-work back to canvas only.


glangford:
The performance is definitely worse, but I don't see profiling information to explain it. I would guess that canvas in general will always be much faster given GPU-accelerated 2D operations and no DOM overhead. 

Issue #28 has the title "Move aircraft tags" but I am wondering if the root of the issue can be solved without making tags draggable by the user?

I assume the driving use case is really when aircraft are ILS locked; in this case aircraft are packed closer together, and on parallel runways the aircraft info blocks overlap each other and the aircraft. With this in mind some options might be:
- make the aircraft info generally smaller to reduce the chance of overlap
- use a special compact info display mode when aircraft are ILS locked, designed to fit within the space defined by the landing separation minima
- use a thin connecting line to the aircraft dot (perhaps only when ILS locked) as implemented in t-data-blocks
- algorithmically place the aircraft info using the landing runway as context. For example, landing on 28L and 28R in KSFO, for 28R info could be placed north-east of the aircraft, while 28L could be south-east so the info blocks are staggered; when zoomed out this feature also enables seeing which aircraft is on which runway.
  A setting for this might be in the airport .json.

e.g. using a screenshot running t-data-blocks, similar to this picture but with more compact info:
nese

tedrek:
Thanks for confirming the poorer performance, I'll try to re-work this using the original canvas drawing method.

For me, the driving use case is keeping clear access to data when there is vertical separation.  Specific places are crossing departures and arrivals and vectoring around for ILS intercept.  It will be very nice to have the information available when they are ILS locked, but the majority of the times I really need it they haven't been assigned a runway yet so I don't see a good way of algorithmically doing it.

glangford:
> For me, the driving use case is keeping clear access to data when there is vertical separation. Specific places are crossing departures and arrivals and vectoring around for ILS intercept.

Understood.

Here is a data point (Euroscope) on tag display and movability, just fyi for comparison. Skip to the bottom. Tags can be moved and also pinned to a screen position.

http://www.euroscope.hu/mediawiki/index.php?title=TAGs

erikquinn:
This is already addressed by #432 in part, and further capabilities like this when different display profiles get added, as described in #716. Closing.

PR #205, opened 2015-10-15
Add Oslo-Gardermoen airport

glangford:
While this is a bit of an obscure choice, Oslo was the first airport in the world to implement a Point Merge air traffic control system and airspace. All ENGM standard arrival routes use Point Merge, and it requires a different approach to manage. It is an interesting variation on the existing airports.

Here is a quote describing the Oslo point merge:

> Each STAR contains segments forming a curved sequencing leg equidistant from the Merge Point (MP). The sequencing leg shall be regarded as a delay manoeuvre for use during periods with heavy traffic. All STAR procedures are described from the start point via intermediate waypoints to the MP, from where an instrument approach procedure commences. Arriving aircraft established on the STAR can expect clearance direct to MP when traffic permits. Succeeding aircraft will subsequently be cleared direct to MP when sufficient spacing to preceding aircraft is obtained. Hence, a precise sequencing can be achieved whilst the aircraft maintain own navigation (LNAV). 

There is a good writeup on the Point Merge approach here:

http://www.roger-wilco.net/point-merge-a-new-art-form-for-streaming-arriving-traffic/

5NM range rings are drawn so that it’s easier to see when to release aircraft to the merge point.


glangford:
@zlsa I had already added nax.json with this pull request prior to #207. However, that file can be removed to resolve the conflict.

zlsa:
I somehow missed this PR completely, sorry. I'll merge it tomorrow.

glangford:
No worries, thanks for the merge!

PR #207, opened 2015-10-19
Adding Norwegian Air Shuttle & Turkish Airlines. Adding LTBA Airport

gitsalty:
Thought it might be helpful to add these two since since they are big European airliners.
Also adding LTBA as well.


glangford:
@gitsalty This is good stuff, a couple of things to point out:
- When an airline is added, the aircraft files must also exist. In nax.json, there is a B733 aircraft, but there is no b733 aircraft json. I sent a pull request to simply delete this aircraft for now since it causes problems for other airports.
- For the LTBA airport, the .json does not get automatically loaded. What you need to do is explicitly load it in airport.js. See `airport_init()`and test to make sure it works.

If you send another pull request with the airport.js change, I'm sure @zlsa will add that in.

PR #208, opened 2015-10-22
Fix nonexistent aircraft in new airline and jshint warnings

glangford:
Fixing 2 issues introduced in #205 and #207


zlsa:
I should probably have tested this but I was in a rush... I think @diastro
has push permissions if this is game breaking.

glangford:
It breaks ENGM but the other airports work ok.

Issue #206, opened 2015-10-18
Adding Istanbul Atatürk Airport (LTBA)


gitsalty:
This airport sounds very interesting especially since there are plans for a new parallel runway (currently 3 runways). Also one of the heaviest traffic movements in the world. I would also love to pull request and add but I'm new to this.

gitsalty:
Also I created THY, and NAX files also I just created LTBA where am I able to pull request cause I'm not able to.

zlsa:
To open a pull request, you have to fork this repository (the Fork button in the top right), then push to your copy as usual. From your repository, you can then open a pull request back to this one.

gitsalty:
Got it thanks!

PR #209, opened 2015-10-22
Update airport.js

gitsalty:
In accordance to #205, #206, and #208. Cross fingers this doesn't break.


glangford:
@gitsalty - almost there, but there are still a few issues to correct. The file format for defining airports is picky, and LTBA doesn't load correctly yet.

If you make these changes then push to your patch-1 branch, this pull request will automatically update:
- Remove the leading zero from the wind angle ("025")
- Remove the blank line that is just after "fixes"

gitsalty:
@glangford Thank you for the advice and help! Fixed the issues mentioned.

glangford:
:+1: 

@gitsalty Glad to help! In future, what you want to do is fully test your changes before submitting the pull request. You can do this for example by simply opening your copy of the project index.html in your browser.

zlsa:
Merged.

PR #215, opened 2015-10-30
EDDM new

harp71:
Everything is new. I screwed uop my githup, so i hope this works. And btw delete all the mess i may have caused :)


harp71:
I had to delete my fork, i don´t know if it was necessary or not, but i had some difficulties and didn´t know how to solve them another way. So from a fresh start i close this now and send a new one. Maybe i get it this time ...

Issue #219, opened 2015-11-05
Terrain, collisions and restricted areas


culebron:
I'd like to see this in the simulator, and here are some tests that I've done. I'm not ready to commit code, but here are some thoughts and sample data and code.

The most interesting airport in terms of terrain, I think is Galeon, Rio de Janeiro (SBGR). It has mountains at north and south, near the coast. Also glideslope to runway 10 is between two hills.

I made some elevation levels rounded to 1000 feet:

![selection_008](https://cloud.githubusercontent.com/assets/119675/10965196/20779b4e-83cd-11e5-8812-7f77af354434.png)

Then simplified geometries, removed too small features and made convex hulls for each feature.

![selection_006](https://cloud.githubusercontent.com/assets/119675/10965213/4715152e-83cd-11e5-93a4-c4f54665a383.png)

Here's a sample of input GeoJSON and resulting terrain files.
https://dl.dropboxusercontent.com/u/6721960/2015/11-05-gig/gig-sample.zip

The idea is to simply check if an aircraft is inside one of the polygons and inside its height range.

To reduce calculations there should be optimizations:
1. The final file is split into levels (305-610m = 1000-2000ft), and we only check one level, in which the aircraft is now. Each level has small number of polygons.
   
   {(0, 305.0): 40,
    (305.2, 610.0): 84,
    (610.2, 915.0): 58,
    (915.2, 1220.0): 28,
    (1220.2, 1525.0): 14,
    (1525.2, 1830.0): 5}
2. Every polygon has bounding box, which we initially check. If ac is out of it, we skip the polygon.
3. Every polygon has a convex hull, which is another polygon, much simpler geometrically (e.g. 300pt polygon reduces to only 13pt)

So, we only have to run the expensive point-in-poly calculation if an aircraft happens to be close to a feature. Though at the lowest level (0-305m) there's one big terrain polygon, which we might need to split to avoid having a big convex hull that matches too many aircrafts.

And finally the biggest optimization is this: calculate ranges to features on current level and track the flown distance to see if the plane may have reached it. If it has flown less than that range, then it just can't be inside that polygon.

Algorithm:
1. When an ac enters some level, we calculate its distance to every polygon and save them as a dictionary.
2. Every turn we substract the flown distance from the saved range for each of them.
3. As soon as one distance becomes 0 or negative, re-calculate the range to that feature again.

[...]

PR #222, opened 2015-11-07
#147 #219 added 'direct' command with shortcuts

culebron:


glangford:
Nice!

A comment on air traffic phraseology - can you update this so that the command is (for example):

"BAW123, cleared direct ADOPI."
with the response
"Cleared direct ADOPI, BAW123"

There is some discussion about the proper phraseology (and it probably varies) but that's my best guess for an ATC direction that gives a shortcut.

/cc: @nielsvdweide @Maverick283

See here:
http://forums.jetcareers.com/threads/proceed-direct-fix-vs-cleared-to-fix-on-stars.120209/

nielsvdweide:
@Glangford 

I think atc phraseology varies in a lot of countries! I've hear many different calls regarding this. I think in general, it's always good to tell someone is cleared to fly direct to a waypoint so I think that's the best way to implement this. I've also heard things like: "fly direct ADOPI or just Direct ADOPI" 

My opinion is that this command should be added with the cleared in front of the command, which will make it clear for everybody to understand!

thanks for the cc by the way! Enjoy your day

culebron:
Ok, will fix when will be at the computer. Tried playing, it was so easier .. Need to add sid/stars, because right now playing sfo or ams is just fix typing contest :)

glangford:
> Need to add sid/stars, because right now playing sfo or ams is just fix typing contest :)

LOL you are right that it is fix typing. I hope to have more time to add SIDs soon.

glangford:
:+1: 
/cc @zlsa

PR #192, opened 2015-09-25
SBGL and SBGR update

MaicomMR:
SBGL and SBGR converted to real coordinate system, added real waypoints
and improved with the real distances between they.

PR #223, opened 2015-11-08
Brasilian airports updated and improved

culebron:
I took the airports files from @MaicomMR #192 and improved their data a bit: added fixes from X-Plane nav data file, changed coordinate system (SXX.XXXX instead of SXXDXXMXX). Also updated airlines and departure destinations (I flew there IRL, and know it a bit).

I did not try to balance rates here.

Embora, galera!


culebron:
Regarding departures, here's the departure to east/southeast from GIG that I flew (CAX YLA PP DACIR)
image2993

I noticed that some fixes on the glideslope are off the centerline. But I think it's better to merge it for now than keep a stale branch.

culebron:
The traffic between RIO and SAO/CWB is so intensive they made 4 parallel airways.

http://airtrafficcontrollerbrand.blogspot.ru/2013/11/nova-tma-rj-partir-de-12-dez-2013.html
chart

culebron:
@zlsa this is ready to be pulled.

PR #231, opened 2015-11-10
Another update of UUDD, FlightRadar24 data

culebron:
In the previous pull request, the arrival directions created about 100 arrivals per hour, which is unrealistic. The real arrival amount is this, and I took 25 flights per hour as a good challenging level.

selection_004

Here's the arrival company breakdown. Aeroflot and S7 are doing some flights for Transaero under its codes.
selection_003


MaicomMR:
Wow, well thought out by using the data from the Flight Radar. When I created the UUDD to help the project did not have much knowledge about the airport, I looked over at wikipedia and algusn other sites, created more to add a Russian airport, since at the time we had few airports options. :+1:

culebron:
@MaicomMR thanks!

Another update: I looked at FlightRadar24 in speed-up mode and saw the arrival and departure patterns. But to be sure I took the arrived flights table from FR24 and analyzed destination bearings to be sure.

Airports database is here: http://openflights.org/data.html
And here is a compass bearing function (no need to write own code nowadays :) ) https://gist.github.com/jeromer/2005586

This gave a table of bearings.
selection_009

selection_008

glangford:
:+1:

culebron:
#237 includes this pull request

PR #238, opened 2015-11-13
Aircraft tails are shown ahead, and proportional to speed

culebron:
If you watch this, you'll notice the "tails" show _ahead_ and are proportional to speed.

[ATC display](https://www.youtube.com/watch?v=qQM6KAnd_rU)

### DIFF
--- a/assets/scripts/canvas.js
+++ b/assets/scripts/canvas.js
@@ -362,16 +362,11 @@ function canvas_draw_aircraft(cc, aircraft) {
       cc.globalAlpha = 0.3 / (length - i);
     else
       cc.globalAlpha = 1 / (length - i);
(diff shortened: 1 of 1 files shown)


zlsa:
This is kind of a large UI change: @glangford @tedrek, thoughts?

glangford:
> This is kind of a large UI change: @glangford @tedrek, thoughts?

For sure it is more correct to have the pointer ahead to emulate real world displays. The video above is from a North American high altitude display (FL240 +). Here is a lower level altitude example (Eurocontrol) that shows the same idea:

https://www.youtube.com/watch?v=w4BOiwqSoB4

In terms of actual game play, it helps to confirm current heading without having to click on the aircraft, and I think people will adjust to it quickly. I ran the show-tail-ahead branch and it worked well for me. :+1:

Issue #235, opened 2015-11-11
[Feature Request] SID show/hide button


MaicomMR:
Hi all, I have been thinking about a SID button to enable or disable the dashed lines and names of SID, take a look on example.

SHOW: ![Alt Text](http://i.imgur.com/lqahRpk.png)
HIDE: ![Alt Text](http://i.imgur.com/cEDYdDz.png)

zlsa:
I think it should be one of the buttons in the bottom row. Any ideas for a smaller icon? Maybe just "SID" or maybe an arrow or something?

culebron:
Probably should look like an aircraft taking off icon which is used in airports.
![download](https://cloud.githubusercontent.com/assets/119675/11109593/e9a9a084-8913-11e5-8ea9-d22df5ce09fa.png)

MaicomMR:
Here is a new test:
On bottom bar [test icon 1]: ![Alt Text](http://i.imgur.com/vny3G7Q.png)

icon[icon 1]: ![Alt Text](http://i.imgur.com/KkXH4VX.png)

On bottom bar [test icon 2]: ![Alt Text](http://i.imgur.com/6t1IQfO.png)

icon[icon 2]: ![Alt Text](http://i.imgur.com/liI4AWx.png)

glangford:
The airplane taking off icon looks great @MaicomMR and @culebron - another possibility would be to create a button image for a group of settings (like a gear icon). @tedrek was experimenting with that idea.

That could contain additional things in the future like Variable wind (on/off), and any other options. eg.
https://commons.wikimedia.org/wiki/Category:Black_gear_icons

I like the look of the airplane taking off the best, but I'm not sure which is better long term.

culebron:
If you ask my opinion, I would put this button on the left, and next to it I'd put all actions/switches that are related to ATC activities, and may be needed often. I guess ATC may quite often want to switch between clean view and "cluttered".

What's important is that these switches should be easy to switch back and forth without irreversible or expensive consequences. (E.g. change airport is irreversible.) STARs will come too, and should also be near SID toggle.

Other buttons are meta-game functions (change airport), actions that don't happen often but still need to be visible and easy accessible: multiple airports are what attracts a newcomer.

culebron:
In #240 I added both buttons to toggle SIDs and restricted areas. Check that out :) Open that branch and see how it works. The canvas are redrawn not immediately, and this gives it a special analog device feeling. :)

[...]

PR #240, opened 2015-11-14
Restricted areas, vector functions, SID/restricted UI switches [needs test]

culebron:
#219 **UPDATE**: Here's the final version, test it please.

The branch is deployed here for tests:
http://culebron.github.io/atc/

For some reason, arriving aircrafts start climbing higher right after spawning. Can someone investigate? I've checked and can't see the reason.


MaicomMR:
Man, u are a genious! Thanks for this :smiley:

culebron:
@MaicomMR thanks! That's just geometry, I like it. :)

@zlsa @tedrek @glangford @nielsvdweide @harp71  please try this out on UUDD or your airports if you wish to add restricted areas.

culebron:
I added both restricted area and SID display switches requested in #235.

culebron:
Made some tests and adjustments. As I found out, bounding boxes did not cure the problem very well. Instead they often caused more calculation cycles when planes flew around a restricted area, touching the bbox boundary.

Also, there were serious CPU spikes when planes entered the area at shallow angle. I made it calculate the range not more often than every 1 km.

selection_046

Also, tried bbox calculation with 1km trick, but it did not help. In fact, using bbox caused higher CPU load. So, I fixed a problem that arised with bbox, and that's enough.

glangford:
This is looking good @culebron , a couple of observations and suggestions:
- the rendering of the restricted areas is nice
- should there be a bigger penalty for going through a restricted area? Right now it is lower than missing the departure window.
- UUDD has a STAR route with fixes ITINA, RW. This passes directly through restricted area UUR-68, but arriving aircraft never received a penalty passing through there. (?)
- should any STAR go through the restricted area?
- this branch has the change to add buttons for enable/disable of SID display and restricted areas; should these changes be a separate pull request?
- when restricted areas are disabled, a bright red bar is displayed over the icon which is distracting. When SID display is disabled, nothing is displayed over the icon which I think is better.
- arriving aircraft are slow (200 knots)

culebron:
> should there be a bigger penalty for going through a restricted area? Right now it is lower than missing the departure window.

Ok, will raise it a bit.

> UUDD has a STAR route with fixes ITINA, RW. This passes directly through restricted area UUR-68, but arriving aircraft never received a penalty passing through there. (?)

That's correct, because that area works from ground up to flight level 20 (200 feet). I'll try making the display similar to charts, so that the height range is clear.

> should any STAR go through the restricted area?

...and hence that STAR does not cross the area.

> this branch has the change to add buttons for enable/disable of SID display and restricted areas; should these changes be a separate pull request?

Ok, I can separate that branch.

> when restricted areas are disabled, a bright red bar is displayed over the icon which is distracting. When SID display is disabled, nothing is displayed over the icon which I think is better.

That was intended, since turning them off is less safe, hence incentivise to turn them on. If this is really so needed and so disturbing, we can remove this red indication. **What do others think?**

> arriving aircraft are slow (200 knots)

I checked with flightradar, you're right, they arrive at about 240kts, will increase the speed.

culebron:
All items mentioned by @glangford  are fixed. Please write your opinions on the UI toggle. Should it be red when restricted areas are hidden, or not?

culebron:
Test the branch here: http://culebron.github.io/atc/

PR #241, opened 2015-11-14
EDDH Improvement

Fechulo:
I thought EDDH was lagging behind the other airports, so I added SID's and STAR's. I based my changes off of these charts: "http://nav.vatsim-germany.org/files/edww/charts/eddh/public/EDDH_AIO.pdf". I slightly modified some of the STAR's to make things quicker.


culebron:
If you set runway ends, it's better to remove heading.

Right now the program code will turn it, and heading even in 0.001s of degree is not precise enough, the localizer may miss the approach markers.

Fechulo:
I have now removed the runway headings.
Thanks for the recommendation!

Fechulo:
I tested the new version and saw that there were many problems, so I fixed most of them. However, there seems to be a problem with the position of the runways. I have tried using coordinates from many different sources but none of them seemed to align properly with the fixes. The fixes seem to be in the right position, so i don't think that's the problem. Any suggestions on how to get the perfect coordinates for the runways?

Fechulo:
Still not perfect, but getting closer.

glangford:
Another possible source of information:
http://airportnavfinder.com/airport/EDDH/#rw

It says for example that runway 5 is at
53° 37' 5.63"N 
9° 57' 49.36"E

which is a bit different (and has more decimal places :smile: )

Fechulo:
Still not perfect, but thanks for for the help

glangford:
> Still not perfect, but thanks for for the help

The locations of the fixes themselves are also not accurate. If you update the runways according to the link I gave, and use opennav for the fixes, you will find they line up much better.

For example:
http://opennav.com/waypoint/DE/DH065

glangford:
eddh

Fechulo:
Alright I'll see how OpenNav works. I've been using the coordinates in the charts so far or skyvector.

PR #243, opened 2015-11-15
EHAM various improvements (should this pull request be closed?)

Fechulo:


glangford:
Can you test locally if you haven't done so already?

Fechulo:
I tried testing it manually, like you told me. However it never loads. I thought it might be a problem with Chrome, but it didn't work in Internet Explorer either. Any suggestions?

culebron:
Opening it locally won't work, because json files are loaded over ajax, which does not work from file path (to prevent reading files from user's computer).

You can use a small http server https://www.google.ru/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8&client=ubuntu#newwindow=1&q=tiny+http+server+windows

glangford:
> Opening it locally won't work, because json files are loaded over ajax, which does not work from file path (to prevent reading files from user's computer).

It works fine for me. I test changes all the time just by opening locally (Safari on OS X).

Fechulo:
It didn't work in Safari either, I'll have to check what @culebron suggested. Which tiny sever would you suggest?

culebron:
@Fechulo  there are several of them,
https://www.ritlabs.com/en/products/tinyweb/
http://sourceforge.net/projects/miniweb/

zlsa:
I know that Chrome won't allow `file:` AJAX; I think Safari and Firefox are more lenient. In either case, it's probably better to self-host it.

(If you have python3 installed, you can run `python3 -m http.server` in the atc/ directory and it'll host that location as `localhost:8000/`.)

Fechulo:
I hosted it locally, but sometimes it didn't load and when it did load it didn't work properly. Even though I have Python 3.5 installed, when I execute a py command, it says that it is using Python 2.7.9

Fechulo:
I was away for a few days, however just before my departure, I did manage to host it locally, test it and fix the bugs. I came back today and saw that another pull request regarding improvements to EHAM had been merged. This new version has more or less the same changes as the one I created, so I don't see any point to keep this pull request open. What do you think?

PR #249, opened 2015-11-16
EIDW loaded

Fechulo:
How can I open a single pull request with multiple branches?


culebron:
You've added `airport_load`, but EIDW file is not in the tree, you need to add it (`git add`) and commit.

glangford:
> You've added airport_load, but EIDW file is not in the tree, you need to add it (git add) and commit.

You can just `git add` the file to your patch-7 branch - when you do that, the pull request is automatically updated.

culebron:
@Fechulo looks like you created one pull request per commit, no need for that, just make one branch and one PR, and push into that branch. Only if you work on separate tasks it makes sense to have several branches and requests.

Fechulo:
Thanks for the help! Unfortunately, I'm a complete noob, so where exactly do I put the "git add" command.
Thanks!

PR #251, opened 2015-11-16
Mountains rendered on canvas, collisions

culebron:
Contunuation of #250, proposal: #219 
@MaicomMR @glangford @nielsvdweide @harp71 
Try this out here: http://culebron.github.io/atc/

selection_070

selection_067


culebron:
For some reason, the message that a plane has crashed disappearrs immediately. (I guess because I make altitude=0 to make it be removed quickly.) Can someone help with this?

zlsa:
@culebron, there's a `crashed` flag in the aircraft class. Setting `altitude` to 0 will instantly remove the plane; setting the `crashed` flag will automatically stop the plane and slowly lower it to the ground.

culebron:
@zlsa thanks! Will check the flag. The problem is that when it collides with terrain it's already on the ground. It would look like it's falling into it. Will look into that code more, see what I can do.

culebron:
I checked what was wrong there. Messages disappeared quickly because I changed CSS style for `.hidden`. It was used for fadeout, I by mistake forced this class to be invisible.

The property is called `hit`, not `crahsed`, and it worked just as I wanted it to: when aircraft collides with terrain, it is already landed, and may disappear. It shouldn't fall down through the ground. So, the fix is pushed.

I think there are no more improvements to push here, unless someone finds a bug.

culebron:
I'm making an improvement: rather than custom JSON, I want to use GeoJSON, which anyone could create with an editor.

culebron:
Major update.
- terrain file is GeoJSON, you can create it with tools like
  - [SRTM2OSM](http://wiki.openstreetmap.org/wiki/Srtm2Osm) and [JOSM](https://josm.openstreetmap.de/)
  - [QGIS](http://www.qgis.org/en/site/)
- Relief lines can be multipolygons now, **with holes**. They are rendered correctly.
- terrain in Rio de Janeiro is much smoother (used buffer-debuffer-buffer method, rather than line simplification algorithms)

selection_009

selection_010

culebron:
Two screenshots: 1) plane flying in a hole in the polygon

selection_013

2) collision in the right place
selection_014

culebron:
Made fixes readable on the map when terrain is displayed.
Test it: http://culebron.github.io/atc/

selection_017

culebron:
To make aircrafts not collide with mountains early on arrival, I arranged STARs for arrivals.

culebron:
...and added SIDs.
Try out: http://culebron.github.io/atc/ (to clean up cached files, open Developer Tools in Chrome - F12, go to options (gear icon) and set "Disable cache")

culebron:
Forgot one thing: the scale :) Will add soon.

zlsa:
Is this ready for merge yet? It looks amazing :)

culebron:
Not yet. I need an elevation scale and sort SIDs at Galeão.

culebron:
@zlsa now it's finalized, you may merge `mountans`.

zlsa:
Thank you very much. This is incredible work.

glangford:
It looks amazing. :star2: :star2: :star2: :star2: :star2:

culebron:
@glangford @zlsa thanks!

Maverick283:
Great work!!!!

Is there any way to tell how high the terrain actually is?

culebron:
I'm going to add a legend key.

Maverick283:
Awsome!!!!

Is this going to be autogenerated or does it need to be done for every airport manually? ;)

harp71:
Nice, but: which color represents which altitude? Shouldn´t there be a reference which color is wich minimum altitude?

culebron:
@harp71 there is a pull request for that, see the legend key shown here: http://culebron.github.io/atc/

@Maverick283 It has to be made manually. But I checked the implemented airports, and only LAX, MUC and DUB have significant terrain around. DME has no hills higher than 250 m (above the airport). Among other world busiest airports, only Barcelona (BCE), Taipei (TPE) and Hong Kong (HKG) have mountains that affect traffic. Madrid Barajas (MAD) has mountains 6000' above airport, but 40 km far. Even Guarulhos (GRU) that we have here has hills just 1000' above the airport.

I want to make a short instruction how to prepare terrain file.

PR #253, opened 2015-11-16
New Airport: EIDW (Dublin)

Fechulo:
somehow, I finally managed to put all the changes into one single pull request. I based all of the SID's STAR's and other details off of these charts: http://www.iaa.ie/safe_reg/iaip/aip_eidw_charts.htm . I don't know how to test it, so hopefully it works!


glangford:
A couple of fixes are required:
- eidw.json needs a "{" in the first line. `{ "name": "Dublin Airport",`
- In a similar way that the airport is loaded, you also need to load the airlines. See airline.js and load Aer Lingus and Ryanair there

Note that you can test locally by opening your copy of index.html.

glangford:
Here is how it looks:

eidw

Fechulo:
thanks for the info and the picture, looks good to me!

glangford:
:+1:

PR #260, opened 2015-11-23
Added social media meta tags

culebron:
See this one working at http://culebron.github.io/atc/


zlsa:
Thanks! This is something I'd intended to do for a long time now but I haven't had the time.

PR #264, opened 2015-11-24
Virtual fixes

culebron:
As suggested by @glangford , I made fixes with underscore prefix not show in the chart. Now you can create much nicer departures with them.


culebron:
Oops, this branch contains too much garbage.

PR #266, opened 2015-11-25
Added "proceed" command

culebron:
This command adds more fixes after current path. Very convenient for planes arriving via STAR. Much less typing fest.


zlsa:
This means that:

```
aircraft fix foo bar baz
```

is the same as

```
aircraft fix foo
aircraft proceed bar baz
```

right?

zlsa:
There's a bug with this PR:

```
/atc/assets/scripts/aircraft.js:712 Uncaught ReferenceError: i is not defined
```

https://github.com/zlsa/atc/pull/266/files#diff-6dbd61076955b1ac77a4a2df790178efR712

culebron:
Yes, it will be the same.

Fixed the bug.

Issue #261, opened 2015-11-23
Parallel Approach


Maverick283:
I was playing at KSFO the other day, and had a huge wave (lots of fun, thanks! :+1: ) coming in. 

As they were coming in from two sides, with their separation being minimum already, I wanted them two land at two runways simultaneously. As traffic got closer to the extended center line, the white circles started popping up, but as I cleared them on the approach (will be creating a new issue about that as well), they would disappear (great job!)

Here is the thing though:
When having two airplanes on a parallel approach (one for 28R and the other one for 28L), you still get the red circles and some points deducted.
KFSO (and many other large airports) allow for a parallel  approach. [(Page 12 for more detail)](http://www.faa.gov/documentLibrary/media/Notice/N8260_73.pdf)
This means that two aircraft can fly almost side by side as long as they are on the ILS and the later traffic has the one in front in side. 

For simulation purposes I think we could skip over the "in sight" rule, but it'd be great if you could have two planes land parallel, as it gives you the opportunity to get traffic in faster, and handle even more traffic!

glangford:
Good stuff. I set up the arrivals for KSFO so that both 28L and 28R are needed for landings at peak.

There are no special rules in the simulation for SOIA or SCP, however it is possible to play with the current KSFO  arrival rate without point deductions. What you need to do is vector the aircraft for landing slightly staggered. This sometimes means sending some of the BIGSUR2 arrivals straight north, and joining the flow with a timely vector to PONKE, for example. Then you can interleave the two major sources of traffic.

culebron:
Just keep one altitude, say 4K for one side and another one, 3K, for the other side. This way there's little risk of colliding them if you forget to give a vector on time.

Maverick283:
@glangford Yeah, thats what I did after I figured out that I would get less points for separation violation. 
What I want to achieve here is that when a plane is on the approach (a.k.a. cleared to land in this simulation, check out #262 ) another plane can join the parallel approach without violating the separation minimums. 

If there is a higher rate of arrivals so that you more or less have to use parallel approaches (which I would welcome!!) then this should not result in a lower score.

@culebron That would work for the initial approach phase, but as soon as both planes start to descent with the glide slope, the same problem will occur.

culebron:
I played SFO again, and indeed, on short final they shine red if going one next to another on different runways.

culebron:
I created a pull request with a penalty for close calls (#267), and will need to take this into account.

tedrek:
[...]

PR #267, opened 2015-11-27
Warnings are penalized [wip]

culebron:
Right now warnings (close approaches) are not penalized. I added this scoring, but right now it also penalizes parallel approaches in KSFO (#261). This has to be fixed.

You may try the new behaviour at http://culebron.github.io/atc/

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -1348,22 +1348,22 @@ var Aircraft=Fiber.extend(function() {
         if (this.isPrecisionGuided() && other.isPrecisionGuided()) {
           if(this.requested.runway != other.requested.runway) {
             // Notice at 3500 feet horizontal and 1500 feet vertical
-            if ((distance < 1.067) && (altitude < 1500)) notice = true;
+            notice = ((distance < 1.067) && (altitude < 1500));
             // Warning at 3000 feet and 1000 feet vertical
-            if ((distance < 0.914) && (altitude < 1000)) warning = true;
+            warning = ((distance < 0.914) && (altitude < 1000));
           } else  {
             // Notice at 2.8nm horizontal and 1500 feet vertical
-            if ((distance < 5.2) && (altitude < 1500)) notice = true;
+            notice = ((distance < 5.2) && (altitude < 1500));
             // Warning within 2.5nm horizontal and 1000 feet vertical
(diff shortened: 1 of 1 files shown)


Maverick283:
This only means that the "red circles" are penalized, but the white ones aren't, right?

Code looks like you already fixed the parallel approach issue?

culebron:
No, not fixed yet.

Maverick283:
Hm, reading this wrong then.

Wouldn't a simple condition checking for the ILS locking status of an aircraft do the trick? As in:

If (!isLockedOnILS) {
    //Do the whole speration thing
} else {
   //aircraft is immune to any separation calculations, maybe needs an additional variable
}

culebron:
This is too simple, because this will permit situations like in the picture: one or both planes come from the wrong side and get ILS locked in the corridor triangle, still on the wrong side, without triggering a penalty.

selection_030

I think we need to check and make an exception (not trigger a warning) if both planes are landing on a parallel rwy and have separation >300 ft. But this still permits situation like in the picture. I think we need to add one more condition: planes should be on different sides of the glide paths (or middle of the 2 glide paths).

Maverick283:
Okay, say we have aircraft 1 (a1) and aircraft 2 (a2). a1 is already on the glidepath, e.g. for rwy 09L. 
The rule is that a aircraft may only fly closer to another aircraft when it has it insight, only then it may start the parallel approach. For simulation, we could adopt this and say that the aircraft may only close in from the side it will fly its approach at. 

So with a1 being on the approach, a2 may come closer to a1, but may not come any closer then the distance between the two runways (here also the approaches). 

This would require to be able to tell how far a aircraft is to the right/left of another aircraft. Then you would have to know the distance and whether one approach is to the right or left of another. 

This is the solution with the best efficiency that I can come up with right now....

tedrek:
#294 has a significant re-factor to the whole notice/warning system but doesn't include the warnings being penalized.

If the reduced lateral separation of 600 feet (or so) is only in place while the aircraft are within say 75 feet of the centerline, then the crossing case will be correctly penalized if there is not sufficient vertical separation at the time.

The only other piece I think is missing is setting flags on the SFO runways in particular to allow such a separation.  Pulling the same stunt at LAX should be a major no-go.

erikquinn:
All issues discussed here were addressed and resolved in a recent merge (#426).

PR #274, opened 2015-12-01
Key for terrain elevations

culebron:
Check it out here: http://culebron.github.io/atc/
(SBGL or KSFO)


zlsa:
Can you offset the key by 0.5 pixels?

culebron:
Did you mean add .5px between the boxes? See at http://culebron.github.io/atc/

zlsa:
No, I meant offsetting x and y by 0.5px (with `cc.translate(0.5, 0.5)`). (On my non-HiDPI display, the edges are on a half-pixel boundary, so they're blurred and 2px wide.)

culebron:
Ok, did that. The text was also blurred, I fixed it as well.
before and after:
selection_010 selection_012

PR #281, opened 2015-12-09
Add Option to make range rings user defined, per airport, via a new property in each airport.json

erikquinn:
First, let me say what an incredible job you all have done with this simulator! It has come along very nicely, and I've recommended it as a practice tool for many of my students (I am a Lab Instructor in a collegiate air traffic control program), as this project is in many ways just as useful as the multimillion-dollar ATC simulation software we have in our lab... So well freakin' done! I've been very impressed and would like to provide some of my own input as well, starting with the below improvement:

---

image
image

---

As shown in the images above, range rings of a definable radius (via airport.json) can now be implemented for each airport. If the loaded airport does not have a "rr_radius_nm" property, then will use the old method of four rings within the "radius_ctr" range. In the future, I'd like to add a customizable location as well, as there are some cases where you may want to place the rings somewhere other than on the airport center.

Being from the air traffic control perspective, the first thing I noticed when playing this WONDERFUL game is that it was difficult to quickly get a sense of scale along the final approach course... in my opinion, using a standard of 5.0nm range rings would make it quite a bit easier to get a sense of scale (as that is an established standard RR radius). These changes would make the process of adding/modifying range rings both smarter and easier.

Upper: USING OLD RANGE RINGS (no change to json file)
Lower: USING 5NM RANGE RINGS (set in ksfo.json... "rr_radius_nm = 5")

Feel free to give her a test at [http://erikquinn.github.io/atc](http://erikquinn.github.io/atc), and/or hit me up with feedback.

### DIFF
--- a/assets/airports/ksfo.json
+++ b/assets/airports/ksfo.json
@@ -5,6 +5,8 @@
   "icao": "KSFO",
   "magnetic_north": 13.7,
   "ctr_radius": 80,
+  "rr_radius_nm": 5.0,
+  "rr_center": ["N37.6195", "W122.3738333"],
   "has_terrain": true,
   "position": ["N37.6195", "W122.3738333"],
   "wind": {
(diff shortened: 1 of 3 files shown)


zlsa:
Thanks a lot!

PR #284, opened 2015-12-15
Expose and rate airport difficulty levels

tedrek:
This gives some indication of how hard a given airport will be prior to playing it.  The difficulties are based on the algorithm in #156, it is to be expected that some will need to be modified after further play testing.

### DIFF
--- a/assets/style/airport.css
+++ b/assets/style/airport.css
@@ -32,8 +32,17 @@
 #airport-switch li {
   list-style-type: none;
 
-  padding: 5px 10px;
+  padding: px 10px;
   margin-bottom: 2px;
+}
+
+#airport-switch li .difficulty {
+  display: inline-block;
+  width: 3.5em;
+  font-weight: bold;
+  font-size: 120%;
+  margin-right: 0.25em;
+  text-align: center;
 }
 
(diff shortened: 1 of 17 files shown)


culebron:
No hard airport? Now we urgently need Dubai and Hong Kong! :)

tedrek:
LTBA is still rated as hard.  KMSP and KSRA are also rated as hard but I believe they need some attention.

For expert I've been planning on Chicago O'Hare with something like 100 arrivals/hour.

culebron:
Wow, didn't notice Stanbul was made so hard. I saw it has not so much traffic, has it? Minneapolis isn't high in chartrs either.

I saw Atlanta has 113 arrivals/hour at peak time. But they have just 5 parallel runways, which is easier to manage than O'Hare's 3 different angles.

assets/style/airport.css
```diff
@@ -32,10 +32,19 @@
 #airport-switch li {
   list-style-type: none;
 
-  padding: 5px 10px;
+  padding: px 10px;
```
culebron: syntax error

tedrek:
Thanks for the syntax catch.  Thanks for the merge.

PR #287, opened 2015-12-17
Realistic airspeeds

tedrek:
Add an option to use simplified airspeeds.  By default it is enabled and game behavior is unaffected.

When disabled, aircraft ground speeds are affected by altitude.  The model is 1.6% increase in ground speed per 1000 feet altitude.  The speed shown in data blocks on the radar is always ground speed.

The 1.6% is based on http://williams.best.vwh.net/avform.htm#Mach with a slight increase to account for compression effects.  I didn't do the math to see how much of an effect compression would actually have at the speeds generally encountered in game.

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -1313,6 +1313,10 @@ var Aircraft=Fiber.extend(function() {
 
       var angle = this.heading;
       var scaleSpeed = this.speed * 0.000514444 * game_delta(); // knots to m/s
+      if (prop.game.option.get('simplifySpeeds') == 'no') {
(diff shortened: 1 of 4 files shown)


erikquinn:
Ooohhhh I like this... that will certainly do a lot to make it "feel right"!

@zlsa Going further with this, are the aircraft currently affected by the wind settings? They don't appear to be, based on what I can see. Being able to crank up a stiff 30-40 degree crosswind throws off all your timing down in the final corridor and really makes you think harder as each base leg could differ by 20+ knots groundspeed.... allowing wind to "push" the airplanes would be a really great feature!

tedrek:
Thanks :smile:

Aircraft are not affected by wind at this point, see #289.

PR #289, opened 2015-12-19
Include wind in aircraft movement

tedrek:
Include effect of wind on aircraft movement while simplified speeds are disabled.

As a simplification winds are assumed to increase in speed by 2% per 1000 feet.

The pilots do not compensate for wind, this means tracks to fixes will be slightly curved and the approach track will be slightly off center.  I have not tested ability to land in a high crosswind but they shouldn't be doing it anyway...

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -81,6 +81,7 @@ var Aircraft=Fiber.extend(function() {
       this.heading     = 0;
       this.altitude    = 0;
       this.speed       = 0;
+      this.groundSpeed = 0;
       this.ds          = 0;
 
       this.radial      = 0;
@@ -1316,9 +1317,28 @@ var Aircraft=Fiber.extend(function() {
       if (prop.game.option.get('simplifySpeeds') == 'no') {
         // Calculate the true air speed as indicated airspeed * 1.6% per 1000'
         scaleSpeed *= 1 + (this.altitude * 0.000016);
+
+        // Calculate movement including wind assuming wind speed
+        // increases 2% per 1000'
+        var wind = airport_get().wind;
+        var vector;
+        if (this.isLanded()) {
+          vector = vscale([sin(angle), cos(angle)], scaleSpeed);
+        }
+        else {
+          vector = vsum(vscale([sin(wind.angle + Math.PI), cos(wind.angle + Math.PI)],
+                               wind.speed * 0.000514444 * game_delta()),
+                        vscale([sin(angle), cos(angle)], scaleSpeed));
+        }
+        this.ds = vlen(vector);
+        this.groundSpeed = this.ds / 0.000514444 / game_delta();
+        this.position = vsum(this.position, vector);
       }
-      this.ds = scaleSpeed;
-      this.position = vsum(this.position, vscale([sin(angle), cos(angle)], scaleSpeed));
+      else {
+        this.ds = scaleSpeed;
+        this.groundSpeed = this.speed;
+        this.position = vsum(this.position, vscale([sin(angle), cos(angle)], scaleSpeed));
+      }
 
       this.distance = vlen(this.position);
       this.radial = vradial(this.position);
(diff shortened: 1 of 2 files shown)


erikquinn:
^NICE. Thanks tedrek!

Maverick283:
I like the introduction of wind affecting aircraft! In the future, this should be one of the functions the user can toggle on/off in the options menu. 

The behavior you described the aircrafts to do is called "homing". Yet this is a rather amateur like behavior. I'd like to suggest that planes maintain a crab angle. 
For the simulation, the only affect the wind would have on the aircraft is the speed, which yet again leads to the discussion about what speed is actually reflected on the scope.   

For real ATC, when they give headings and speeds, pilots do not recalculate them to cancel out the wind but strictly follow the instructions. So a plane that was told to head 180° might actually fly 200° if there is a wind from the east. Also, if the plane is told to fly with a speed of 200kts, it might actually be flying 300kts if the aircraft is high enough and has (quite a bit of) tailwind. 

This behavior could be easily implemented I think and would be a good addition to this introduced feature! :)

erikquinn:
^Agreed. The latter (para. 3) IS the current behavior of the sim, in both cases. (See the below screenshots, where a ludicrous 80kt west wind blows an aircraft well off its assigned heading of 180, and another where two aircraft assigned 250 knots show a 160kt speed difference across the ground). Also, it is toggle-able via the "Use Simplified Airspeeds" option. All seems to be in order there.

The only remaining bit of funny business would be the way the aircraft follow their flight plans. Options to address that are either:
 (a) Have all aircraft in "fix" navmode ignore the wind's affect on position, via if-elsing those a/c out of "this.position = vsum(this.position, vector);"
 (b) Make aircraft smart enough to make the heading they face not be the heading that points to their next fix, but rather the heading that their motion plus the push of the wind is combined to point to the fix (this would be the crabbing approach)

b is well preferable to a, however, depending on the amount of realism we want to build in, this change is what we want for large aircraft that have "the good stuff" in the cockpit, but the behavior we currently have (crooked courses to fixes) MAY BE CORRECT for a/c with crappier equipment that doesn't do all the hard thinking for the pilots, unless the pilots are calculating the crab angle correctly themselves. Perhaps incorporating [equipment suffixes](https://flightaware.com/about/faq_aircraft_flight_plan_suffix.rvt) into the aircraft object is a good idea? Probably a bit over the top, but it would be neat to see a tiny guy like a piper cub or something that is /W get pushed around a bit by high winds while the well-equipped airliners fly straight.

image
image

Also, the below runway landing behavior shouldn't be a thing. Flying from fix to fix may yield some crookedness, but neither pilots nor controllers would allow a descent below MVA/MSA unless truly established on the localizer. The pilots would certainly be crabbing here. The very bottom diagram shows what we want (left) and what we have (right).
image
diagram

Maverick283:
That's some good stuff up there, thanks for those screenshots. Didn't see the toggle button yet. ;)

I think we have to differentiate between "heading" and "fix" mode. When an aircraft is given a heading it should fly that heading, making it fly a different course (which is the heading plus whatever the wind does to the aircraft, ignoring magnetic and true course/heading for a second). On the scope, the aircrafts "tail" should not point opposite to the heading as a radar can not pick up the heading of a aircraft but opnly its course. 

In fix mode, the aircrafts should fly directly to the fix, no matter what winds are present. I think it is fair to say that every IFR Pilot is able to do this with a simple Garmin500 (or whatever other GPS) in his cockpit. 

IF we start to give equipment suffixes to aircrafts, we'd have to take into account that some aircrafts might not be able to fly to some way-points. Also, most airliners do have the equipment to have the autopilot fly the direct course from one waypoint to the other, regardless of the winds present. If we later add VFR traffic and/or smaller airplanes on a regular basis, we might consider having planes that (e.g.) do not have a GPS and are unable to fly to GPS waypoints. 

About that last picture: I haven't looked into the code, but in general a simple "mode variable" should do most of the magic: If an aircraft is in heading mode, the wind effect must be added to the heading. If it is not in heading mode (a.k.a. fix mode/ loc mode/ SID, STAR, etc) the wind should simply be ignored.
This would somewhat project reality on the code: Handling the wind is done in the cockpit when flying from fix to fix, and by the controller when flying headings...

tedrek:
Thanks for the comments :).  I've opened issue #296 to deal with aircraft tracks, can we continue the discussion there? (it's much easier to track)

PR #292, opened 2015-12-20
Airline fleets

tedrek:
When building an airport it can be very useful to select from a reduced set of the aircraft which a given airline flies.  This allows specifying multiple sets of aircraft for a given airline and selecting from those sets for an airport.  AAL and ACA have been updated as examples along with SBGR and SBGL.
### Airline

Add a `fleets` key to the airline and populate it with an object.  There should be a `default` fleet specified (an existing `aircraft` key on the airline should usually be converted to the `default` fleet).  Populate other fleets as necessary, common ones are expected to be `short` and `long` for short haul and long haul respectively. 

Example:

```
{
  "name": "American Airlines",
  ...,
  "fleets": {
    "long": [
      ["B763", 2],
      ["B772", 1 ]
    ],
    "default": [
      ["A319", 1 ],
      ["A321", 1 ],
      ["B738", 6],
      ["B763", 2],
      ["B772", 1 ]
    ]
  }
}
```
### Airport

To select a specific fleet for a given airline use a slash followed by the fleet name.

Example:

```
{
  ...,
  "departures": {
    ...,
    "airlines": [
      ["AAL/long", 40],
      ["AAL/short", 1]
    ]
  }
}
```
## Notes

The existing solution has been to create copies of the airlines named for the region which they are flying into.  This has three drawbacks:
- The airline general information is duplicated.
- These sets are non-intuitive to use outside of the region they were intended for
- Each additional .json file adds to the load time (about 150ms per file for me)

By keeping the fleets in the same file as the parent airline the shared information is actually shared.  The load time is only marginally affected by increasing the number of fleets as most of the delay is in the HTTP request/response cycle.  And various fleets should be more intuitive to share between airports.


zlsa:
I'm not sure why I hadn't thought of this as it seems like a very clean solution to regional branches of airlines. Thanks!

Maverick283:
I was trying to apply this to EDDM, but I started to wonder what the number after a aircraft in a airline.json is all about? Can't be the amount in the fleet (if it is, Lufthansa is all wrong), neither the percentage of this aircraft in the fleet (doesn't add up to 100). So what is it?

tedrek:
The number for an aircraft is the relative weight within the fleet.  In a fleet of two aircraft with weights 3 and 8 one of the aircraft will usually show up 3 times out of 11 and the other will show up 8 times out of 11.

I have generally just thrown the number of aircraft in the fleet in there rather than trying to simplify it further.

Maverick283:
Ahh I see, i figured it was something like that, yet this is quite random xD Thanks for the help

PR #294, opened 2015-12-20
Stateful violations/warnings re-factor

tedrek:
The notice/warning system is unable to generate more than one notice for a given aircraft.  It is also difficult to extend the existing system to deal with closing rate based alerts.

This creates a conflict object for any pair of aircraft which are within a 10km box from each other.  Checks can than be carried out on the pair with consistent pairwise state over time.

Specific things which this will allow are:
- Suppressing alerts for parallel takeoffs when headings differ and separation is increasing
- Suppressing alerts for crossing traffic at 1000 foot intervals which are stable in altitude
- Applying in-trail alerts for landing aircraft on the same runway

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -2,14 +2,19 @@ zlsa.atc.Conflict = Fiber.extend(function() {
   return {
     init: function(first, second) {
       this.aircraft = [first, second];
-      this.distance = distance2d(this.aircraft[0].position,
-                                 this.aircraft[1].position);
+
+      this.distance = vlen(vsub(first.position, second.position));
+      this.altitude = abs(first.altitude - second.altitude);
+
+      this.collided = false;
 
       this.conflicts = {};
       this.violations = {};
 
       this.aircraft[0].addConflict(this, second);
       this.aircraft[1].addConflict(this, first);
+
+      this.update();
     },
 
     /**
@@ -50,6 +55,7 @@ zlsa.atc.Conflict = Fiber.extend(function() {
     update: function() {
       this.distance = vlen(vsub(this.aircraft[0].position,
                                 this.aircraft[1].position));
+      this.altitude = abs(this.aircraft[0].altitude - this.aircraft[1].altitude);
 
       // Check if the separation is now beyond the bounding box check
       if (this.distance > 14.2) {
@@ -58,6 +64,13 @@ zlsa.atc.Conflict = Fiber.extend(function() {
       }
 
       this.checkRunwayCollision();
+
+      // Ignore aircraft below about 1000 feet
+      if ((this.aircraft[0].altitude < 990) ||
+          (this.aircraft[1].altitude < 990))
+        return;
+
+      this.checkProximity();
     },
 
     remove: function() {
@@ -96,6 +109,71 @@ zlsa.atc.Conflict = Fiber.extend(function() {
       }
     },
 
+    /**
+     * Check for physical proximity and trigger crashes if necessary
+     */
+    checkProximity: function() {
+      var conflict = false;
+      var violation = false;
+
+      // Reduced horizontal separation minima during precision
+      // guided approaches
+      if (this.aircraft[0].isPrecisionGuided() &&
+          this.aircraft[1].isPrecisionGuided())
+      {
+        if (this.aircraft[0].requested.runway != this.aircraft[1].requested.runway)
+        {
+          // Notice at 3500 feet horizontal and 1500 feet vertical
+          if ((this.distance < 1.067) && (this.altitude < 1500))
+            conflict = true;
+          // Warning at 3000 feet and 1000 feet vertical
+          if ((this.distance < 0.914) && (this.altitude < 1000))
+            violation = true;
+        }
+        else
+        {
+          // Notice at 2.8nm horizontal and 1500 feet vertical
+          if ((this.distance < 5.2) && (this.altitude < 1500))
+            conflict = true;
+          // Warning within 2.5nm horizontal and 1000 feet vertical
+          if ((this.distance < 4.6) && (this.altitude < 1000))
+            violation = true;
+        }
+      }
+      // Standard separation
+      else {
+        // Notice at 4nm horizontal and 1500 feet vertical
+        if ((this.distance < 7.4) && (this.altitude < 1500))
+          conflict = true;
+        // Violation within 3nm horizontal and 1000 feet vertical
+        if ((this.distance < 5.6) && (this.altitude < 1000))
+          violation = true;
+      }
+
+      if (conflict)
+        this.conflicts.proximityConflict = true;
+      else
+        this.conflicts.proximityConflict = false;
+
+      if (violation)
+        this.violations.proximityViolation = true;
+      else
+        this.violations.proximityViolation = false;
+
+      // Collide within 160 feet
+      if (!this.collided &&
+          ((this.distance < 0.05) && (this.altitude < 160)) &&
+          (this.aircraft[0].isVisible() && this.aircraft[1].isVisible()))
+      {
+        this.collided = true;
+        ui_log(true,
+               this.aircraft[0].getCallsign() + " collided with "
+               + this.aircraft[1].getCallsign());
+        prop.game.score.hit += 1;
+        this.aircraft[0].hit = true;
+        this.aircraft[1].hit = true;
+      }
+    }
   };
 });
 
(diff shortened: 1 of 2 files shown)


tedrek:
I would like to hear from @culebron in regards to interaction with #267 before this is merged.

culebron:
I see the logic of separations is mostly the same, isn't it? I think it's a good start. Please merge this, my changes are pretty cosmetic. I only did 2 things: warnings have sort of state, and are compared with the previous state. Next, I made warnings not be scored if one of the planes is out of control range. (BTW, hit outside of control range is still penalized.)

I tried playing with my changes in that pull request, there are other problems and it needs an overhaul and systematic approach.

My initial thought was to make a centerline or an averade center & direction of 2 runways with vectors (make sum and difference and see what result is longer and take that one as the average, then check if planes are on different sides). But this is still patchy.

Another frequent problem is that perpendicular runways are a problem, can't take off until the first plane is kilometers away, and landing + crossing takeoff will always generate a warning. Parallel runways are a problem too.

Maybe let's make a table of conflict events and normal allowed situations?

I also think we might want to add intentions to the equasion, like on SFO approach, when one of the planes is on crossing course with another one, but is going to intercept localizer.

tedrek:
Yes, the logic at this point should be nearly identical, the only difference being each proximity warning can be scored.

@culebron, for the case of parallel approaches I think it can be simplified quite nicely.  Aircraft must be separated by 3nm horizontal or 1000' feet vertical unless _both_ aircraft are within 10 feet of the localizer center line (instead of simply being within the localizer cone).  The only thing this loses is the ability to line up two aircraft from the outside at the same altitude, but that's not really allowed in practice from what I can tell so should probably be penalized anyway.

In reality at SFO aircraft establish on the localizer at 2500' lateral and I believe a 1000' vertical separation.  After establishment the vertical separation is lost.  It's only after the trailing aircraft has the lead aircraft in sight that they close to ~750' lateral separation (plus a bit of in-trail).  At other airports such as ORD/ATL the 1000' vertical separation is applied during establishment on the localizer as well, but they remain at 2500' lateral throughout the approach.

For departures, reality allows parallel departures so long as assigned headings differ by at least 30 degrees and that heading is established by 1 minute after takeoff with separation increasing (or something very similiar to that).

Perpendicular use is allowed so long as one aircraft has passed the intersection point before the second aircraft is given clearance.  The caveat is that in a crossing takeoff the first aircraft across the intersection must not have rotated yet, or the second aircraft must be able to climb above the flight path of the first aircraft at the crossing point, this is to avoid wake turbulence.

I think we can simplify the crossing cases to: No two aircraft may be approaching the intersection of a runway at the same time, at least one of the aircraft must be past the intersection before the other aircraft may move.  If we wanted to be completely accurate, an aircraft which is cleared to land on a given runway should be considered to be approaching any intersections on that runway (which would be an argument for separate approach and landing clearances).

Finally, ignoring proximity under 400' AGL is realistic and covers normal approach+departure situations (both parallel and proximity).

erikquinn:
@tedrek is right on with pretty much everything. Just a few things:

> In reality at SFO aircraft establish on the localizer at 2500' lateral and I believe a 1000' vertical separation. After establishment the vertical separation is lost. It's only after the trailing aircraft has the lead aircraft in sight that they close to ~750' lateral separation (plus a bit of in-trail). At other airports such as ORD/ATL the 1000' vertical separation is applied during establishment on the localizer as well, but they remain at 2500' lateral throughout the approach.

^ The only ways they're separating IFR aircraft is with radar (3nm/1000') or visual separation. Due to the somewhat goofy arrangement of SFO's runways, their centerlines are too close together to be able to use both of them simultaneously for arrivals. How they can get around that is by applying visual separation instead of radar. Goes something like this: (1) clear a/c A for ILS 28R @3k' (2) put a/c B close enough to a/c A that they report them in sight, still >3nm (3) tell a/c B to 'maintain visual separation' and... (4) clear a/c B for ILS 28L. Now they're both on ILS, and ATC still must put all 28R traffic at least 3nm in trail, but when visibility is good, they can have each 28R airplane being followed by 28L lander that maintains visual separation from the 28R guy. This is why SFO in particular is SO vulnerable to delays in crappy weather: when you can't see, you can't use visual separation, so they're pumping ONE arrival every three miles instead of two.

> For departures, reality allows parallel departures so long as assigned headings differ by at least 30 degrees and that heading is established by 1 minute after takeoff with separation increasing (or something very similiar to that).

^Very similar indeed; all the verbage is right, but it's 15 degrees diverging.

> The caveat is that in a crossing takeoff the first aircraft across the intersection must not have rotated yet, or the second aircraft must be able to climb above the flight path of the first aircraft at the crossing point, this is to avoid wake turbulence.

^This is only for heavy aircraft. The rules are somewhat goofy, but they are: (1) any departing behind heavy departure needs to wait 3 minutes or 5 miles (2) any departing from an intersection of the same runway where a large/heavy departed full length needs 3 minutes (3) Rule #1 also applies to intersecting runways, but only if the first departure rotated before the runways' intersection

> If we wanted to be completely accurate, an aircraft which is cleared to land on a given runway should be considered to be approaching any intersections on that runway

^I disagree here. The rule from the tower is that the departure must be past the intersection of the two runways before the arrival crosses the threshold. So by having an airplane cleared to land on a 10 mile final would trigger an alert if somebody started rolling on the intersecting runway.

> Finally, ignoring proximity under 400' AGL is realistic and covers normal approach+departure situations (both parallel and proximity).

^Sounds fine for now, but eventually we should implement the actual rules of 6,000' and wheels-up on same rwy, and past-the-intersection on intersecting runways before either a departure can start takeoff roll or an arrival can cross the runway threshold.

erikquinn:
Also, eventually, some logic for the "passing and diverging" separation rules would be nice to prevent separation losses from being recorded when <3nm apart but separated by these rules instead. See the 7110.65, section 5-5-7-a-1 (pasted below):

> (a) Aircraft are on opposite/reciprocal
> courses and you have observed that they have passed
> each other; or aircraft are on same or crossing
> courses/assigned radar vectors and one aircraft has
> crossed the projected course of the other, and the
> angular difference between their courses/assigned
> radar vectors is at least 15 degrees.
> NOTE−
> Two aircraft, both assigned radar vectors with an angular
> difference of at least 15 degrees, is considered a correct
> application of this paragraph.
> (b) The tracks are monitored to ensure that the
> primary targets, beacon control slashes, FUSION
> target symbols, or full digital terminal system
> primary and/or beacon target symbols will not touch

tedrek:
@erikquinn, that's some great info.  Departing aircraft needing to be past the intersection before an arrival is over the threshold makes it nice and easy, I had assumed there needed to be more separation.

> > Finally, ignoring proximity under 400' AGL is realistic and covers normal approach+departure situations (both parallel and proximity).
> 
> ^Sounds fine for now, but eventually we should implement the actual rules of 6,000' and wheels-up on same rwy, and past-the-intersection on intersecting runways before either a departure can start takeoff roll or an arrival can cross the runway threshold.

To be sure I'm clear here, it is allowed to give an aircraft takeoff clearance if the previous departing aircraft is at least 6000' down the same runway and climbing, this assumes there is not a complication with wake turbulence(?).  I think the proposed rules for no more than one aircraft moving toward an intersection will cover the latter half.

> ^ The only ways they're separating IFR aircraft is with radar (3nm/1000') or visual separation

In regards to this, I was under the impression SFO had a SOIA consisting of  ILS/PRM and LDA/PRM approaches which is still possible to be assigned.  During the portion of the approach prior to breaking clear of clouds separation is maintained by radar and can be less than 3nm/1000'.  I can't find the order but I think I've got the documentation for the type of approach I'm talking about [here](http://www.faa.gov/documentlibrary/media/order/8260.49a.pdf).

In looking for the documentation I also found [JO 7110.308](http://www.faa.gov/documentLibrary/media/Order/JO7110.308CHG3inc.pdf) which appears to authorize radar separation down to 1.5nm utilizing a staggered pattern for SFO among other airports.

All that said, I think the proposed rules of allowing an aircraft on a different ILS path to be within a distance of something like 700' lateral while both of them are on the centerline will give a feel which most closely resembles the actual approaches flown today.  It's a simplification for not having proper visual approaches and should be enabled on a per airport/runway basis.

tedrek:
I think it will be best to work the individual refinements as separate pull requests rather than trying to mush them all together in this one.  Thanks for the discussion.

avdg:
Hmm I'm getting some errors from here:

https://github.com/zlsa/atc/blob/gh-pages/assets/scripts/aircraft.js#L77 (I wonder if it should be `this.aircraft` instead of `this, aircraft` because removeConflict only takes 1 argument)

```
"ReferenceError: aircraft is not defined
    at k.remove (https://zlsa.github.io/atc/assets/scripts/aircraft.js:77:45)
    at k.update (https://zlsa.github.io/atc/assets/scripts/aircraft.js:62:14)
    at k.checkConflict (https://zlsa.github.io/atc/assets/scripts/aircraft.js:1739:31)
    at aircraft_update (https://zlsa.github.io/atc/assets/scripts/aircraft.js:1982:16)
    at update (https://zlsa.github.io/atc/assets/scripts/modules.js:266:3)"
```

tedrek:
Yep, that's what I get for insufficient testing.  Thanks for the message and apologies for breaking it.

PR #295, opened 2015-12-21
Basic conversion of wiki to jekyll documentation

### LINKED ISSUE #293: Location for documentation
Hi Folks,

The wiki and the code continue to be out of sync with each other.  One of the challenges I've identified is the documentation isn't part of the pull request and therefore I'm not in a position to keep the wiki and deployed code in tight sync with each other.  This leads to either writing the wiki update and it being ahead of the deployed code or (much more commonly) waiting until after the pull request has been merged and then maybe remembering to update the wiki.

My proposal is to include the technical reference documentation within the main source code.  This would replace the majority of the wiki as it stands now and appropriate links can be placed.  It also means code updates and documentation updates can happen at the same time.

I'm thinking of specifically using jekyll and markdown to get the minimum necessary for a set of web pages under something like zlsa.github.io/atc/documentation/

@zlsa, @glangford, @culebron, and other's, any thoughts/concerns?

tedrek:
Here's the basic structure to fix #293.  See documentation/README.md for a quick overview, formatting is the same as the wiki.  All three wiki pages are present.

There is currently an [example deployment](http://tedrek.github.io/atc/documentation/).  It's not very pretty right now but it works.

I don't expect to be able to put much time in the next week so thought it would be good to get the framework up for everybody to look at.


tedrek:
Wow, fast merge.  :+1:  Thanks @zlsa.

zlsa:
You're welcome!

PR #297, opened 2015-12-22
Wind compensation

### LINKED ISSUE #296: Deficiencies in aircraft tracking with wind
This is a continuation from comments on #289 and briefly summarizes the content.

The specific issues are pilots when navigating to fixes point the aircraft directly at the fix rather then compensating for any crosswind and when landing aircraft don't specifically compensate for crosswind to remain on centerline.

Proposed solutions for fixes are:
a) Ignore wind motion for aircraft in fix mode
b) Improve aircraft navigation to maintain a constant track in fix mode

B is strongly preferable to A, although depending on avionics in an aircraft not all aircraft may be capable of crabbing during fix tracking.  This suggests implementing equipment suffixes/capabilities.  Equipment suffixes would mean some aircraft shouldn't be able to navigate to certains fixes because of their type.

Implementation of crabbing means the radar vector/tail should be the aircraft's track, not it's heading as is currently done.

tedrek:
Compensates for wind during fix and ils navigation by adjusting the physics step.  Also changes aircraft tails to match groundSpeed/Track.  Fixes #296.

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -82,6 +82,7 @@ var Aircraft=Fiber.extend(function() {
       this.altitude    = 0;
       this.speed       = 0;
       this.groundSpeed = 0;
+      this.groundTrack = 0;
       this.ds          = 0;
 
       this.radial      = 0;
@@ -1345,11 +1346,13 @@ var Aircraft=Fiber.extend(function() {
         }
         this.ds = vlen(vector);
         this.groundSpeed = this.ds / 0.000514444 / game_delta();
+        this.groundTrack = Math.atan2(vector[0], vector[1]);
         this.position = vsum(this.position, vector);
       }
       else {
         this.ds = scaleSpeed;
         this.groundSpeed = this.speed;
+        this.groundTrack = this.heading;
         this.position = vsum(this.position, vscale([sin(angle), cos(angle)], scaleSpeed));
       }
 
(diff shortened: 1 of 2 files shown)


assets/scripts/aircraft.js
```diff
@@ -1327,17 +1328,31 @@ var Aircraft=Fiber.extend(function() {
...
+          {
+            var offset = angle_offset(this.heading, wind.angle + Math.PI);
+            crab_angle = Math.asin((wind.speed * Math.sin(offset)) /
+                                   this.speed);
+          }
+          vector = vsum(vscale([sin(wind.angle + Math.PI),
+                                cos(wind.angle + Math.PI)],
                                wind.speed * 0.000514444 * game_delta()),
-                        vscale([sin(angle), cos(angle)], scaleSpeed));
+                        vscale([sin(angle + crab_angle),
```
culebron: I think you can just use vturn without a vector. It will turn a unitary vector `(1, 0)` to your angle then. Less typing and typo probability.

assets/scripts/aircraft.js
```diff
@@ -1327,17 +1328,31 @@ var Aircraft=Fiber.extend(function() {
...
+                                cos(angle + crab_angle)],
+                               scaleSpeed));
         }
         this.ds = vlen(vector);
         this.groundSpeed = this.ds / 0.000514444 / game_delta();
+        this.groundTrack = Math.atan2(vector[0], vector[1]);
```
culebron: You can use `vradial` here.

```
this.groundTrack = vradial(vector);
```

tedrek:
@culebron, thanks for the review and suggestions.

PR #299, opened 2015-12-22
Add intuitive keyboard shortcuts that re-work user controls

erikquinn:
Adds what I called "shortKey" commands that are a single string, instead of using spaces for parsing commands vs arguments. Essentially:
   'turn left 250'   -->   '<250'
   'climb 5'   -->   '^5'
   etc...

All old commands are still intact, but I stole control of the arrow keys for use with climb/desend/turn left/turn right. The up/down arrows used to be for command history (recalling a previous callsign without having to click). Remapped those to CTRL+up and CTRL+down.

In total, new commands include:
- '^' for climbing
- 'v' for descending
- '<' for turn left
- '>' for turn right
- 'fh' for fly-heading
- '+' for increase speed to xxx
- '-' for decrease speed to xxx
- '*' for cleared ILS XX approach

I tested it thoroughly to make sure nothing was broken, and it all seems to be in order.

Feel free to try it out: [http://erikquinn.github.io/atc/](http://erikquinn.github.io/atc/)

---

Also updated the tutorial to guide users to use the new commands, and explains how they work. While I was there, I updated a few things and made some small expansions and clarifications.


Maverick283:
As soon as this is merged I will add it to the wiki :)

culebron:
From my experience, I use up key more often than climb, so you make user rarely use arrow keys, but more often use more laborous CTRL+arrow.

I'm also skeptical about shortcuts like `<`. `t 90` is 4 key strokes. `< 90` is 5 (shift + `.`). Are you sure it's easier to use? For most commands there are shortcuts that are rather intuitive: `a(ltitude)`, `t(urn)`, `s(peed)`. `p(roceed)`. Missing a shortcut for taxi.

What about keyboard navigation? User often needs to change scale or switch between planes. Maybe CTRL+arrow is better to make selection jump between planes?

I'd also appreciate a way to repeat a command to another aircraft.

erikquinn:
Well it all hinges on having the arrow keys mappings... you're right in that the format ">250" is no better than "t r 250". It's just that when stringing together a bunch of commands, ">250 v3 -210" is more clear and clean than "t r 250 c 3 sp 210". For me, the spacebar is the enemy. having to hit it even once during a single command takes extra time to reach, and detracts from the ability to work the airplanes. I end up struggling more to enter stuff than I do with working the traffic, which is something I desperately want to resolve. And yes, I didn't cover taxi, nor most of the others. Just the core instructions that you need to be able to blast through (alt, hdg, spd, ils). As for the rest, more things _could_ be mapped to specific keys, but I figure the default commands "fix segul" "sid molen" "taxi 28L" are good enough, as they are less time sensitive.

If many people are heavy up-arrow users to recall the last a/c, I suppose that could be left alone, but I just doubt that having to add a ctrl button to that would outweigh the benefits of being able to use the arrow keys for turns/climbs/descents.

culebron:
Well, you don't need right or left in turns, just headings. Also, better traffic management comes with practice. I had difficulties with DME, now I can land all planes on one runway @2x warp, with mostly proceed and direct commands, and a bit of vectoring.

I must say I'm too expert in typing (also with Dvorak layout) to judge if this is an improvement. Maybe ask someone less experienced to try both ways?

Regarding climbs/descents, I don't really see a difference between them. It's all altitude, and we just give `altitude XXXX` command anyway. It makes much sense in real aviation to cross-check each other for errors. If you say `climb 5000` to a plane that already climbed 10,000, they will reply `already above etc...`. Do we really need this strictness here? (It's a serious, not a rhethoric question, please comment.)

erikquinn:
Well you _say_ right and left when you actually give the commands, so I think having the norm being specifying the direction is better than just using "t 250" personally. If not, the old commands would still work. I'm not trying to attack the old commands: they're perfectly fine. I just want a way to do them much faster, and the arrow keys are both next to the numpad and largely unused, so it seemed a good fit to me. Do you have a different suggestion?

culebron:
What if you make a typo and want to jump to that character and change it? I do this occasionally, and it would be great to capture Ctrl+W. :)

erikquinn:
^That's the one issue I totally agree with. using the L/R arrows means you can't move left/right in the text you're typing. which sucks, for sure... having to go the backspace or mouse is unattractive; maybe a key to clear out all but the callsign?

culebron:
Errr.. and retype everything? :)

`<callsign> taxi 32l a 10 sp 250 sid cheru`

Here's another point, again not an opposition but to think: one-letter shortcut is straightforward and learnable, but how many can remember the concept of arrows in text, `< > ^ v`?

erikquinn:
Well, if it's that long, surely you wouldn't want to do so; backspacing or using the mouse to place the cursor would be faster.. but for smaller commands, such as...

```
  BAW193 >190 v3
```

...to change the direction, sure, I would just bite the bullet and re-do it, which takes well under a second, whereas hitting the left arrow 6 times, backspace once, and shift+, would surely be more annoying than starting anew.

culebron:
You're right, I thought of myself, and I rarely do this. More often I just backspace some chars and re-type them.

But for a usual user, I think if you show a text box, everyone expects it behave conventionally, and if it's same box, but does some tricks, it may cause a lot of annoyance.

I'm out of ideas, let others say their thoughts.

**P.S.** My impression right now is that controlling approaches by fixes is a bit more predictable, with a bit of vectoring in some cases, and a good save on workload would be to add transitions (in simple words, these are ways by usual or GPS fixes from STARs to glideslopes).

tedrek:
I agree, overriding the behavior of the left and right arrow keys is a bad idea.  Text entry boxes have a pretty specific expected behavior.

@erikquinn, I am curious as to how you play?  I personally use neither the numberpad or arrow keys in general use so I'm having a hard time understanding what you are doing.  I'm a touch qwerty user who plays with both hands on the keyboard, aircraft are either selected by the right thumb using a laptop trackpad or the right hand moving to a mouse.  Commands are generally single letters followed by numbers entered using the row above the letters.  Any ctrl sequences are relatively awkward (although I use emacs so have learned to cope).

Also, '+' and '-' are currently mapped to zoom commands.

From the sounds of it you are using the numpad a fair bit, my suggestion would be to develop a numpad centric entry mechanism.  If we have an option for say 'Numpad Terse Entry' which is off by default then + and - can be re-mapped for your use when it is on.  Then I would assign the five non numeric keys to arbitrarily be:
- \* - Flight level/altitude
- \- - Turn left
- \- + Turn right
- / - Speed
- . - land

Implementing an alternate parser for this input would mean there are no spaces necessary so you could enter a left turn to 090 with speed 250 and altitude 5,000 as `BAW121 -090/250*5`.  It would be quick to enter but there are no good mnemonics.

erikquinn:
Alright, there's a lot going on in this thread now... so please allow me to clear up a few things. It seems the value of what I'm proposing is getting buried behind a list of problems, many of which aren't actually problems. I've sorted everything below so we can focus on the legitimate conversation points, because I concede they _are_ worth talking about.

---

**_Non-Issues:**_
- @tedrek '+' and '-' _are_ mapped to zoom, yes, but '+' and '-' are not currently mapped to anything. (It's a semantical issue; i'm talking about the +/- in the numpad, which are different keys from the ones next to the backspace button, which control the zoom). Thus, they are available and pose no conflict to zooming.
- @culebron suggesting instead of vectors "controlling approaches by fixes" isn't something I'd ever do, as that's not how it's done at SFO in the real world. You're right: there are instances where a STAR can hook right into an IAF for an instrument approach, but (at least at SFO), there are no such situations. All airplanes are vectored to the final approach course or, at most, cleared to ARCHI (IAF@28R) or MENLO(IAF@28R) with an ILS approach clearance, but depending on the angle they hit ARCHI/MENLO from, they might have to do a procedure turn, which we don't simulate here. If they're vectoring these airplanes in the real world, in my opinion, anything else defeats the purpose of "simulation".
- @culebron the issue of strictness regarding "^5000" vs "v5000" is a good question... I would say if we could hear the pilots, I would want them to reject a climb to a lower altitude, because that would indicate to me that maybe I wasn't talking to who I thought I was. But, since there's no voice (and I don't exactly read the responses very often), it's probably better to allow climb/descents to be ambiguous with regard to direction, and just go to the altitude whether it was a 'climb' or 'descend'. This is the current behavior in the sim, and this pull request leaves it that way.

**_Actual Potential Issues:**_
1. seeing "^ v < >" might be less clear than "a(ltitude) t(urn)", as @culebron suggested. (Personally, I think the nature of them being arrows makes it both more clear and more intuitive)
2. text box would not behave like a simple text box, as the arrows type stuff instead of moving the cursor
3. users who depend on up/down arrows recalling recent callsigns will have to instead click Ctrl+up or Ctrl+down

**_Additions:**_
- ability to clear out last command typed, or maybe just all except callsign (in order to correct a mis-typed command that you can't left-arrow your way over to)

---

**_My Comments:**_
@tedrek yeah, it seems like the reason this isn't being as well-received as I would have thought is that I assumed _everybody_ would be using the numpad. They're all I use; I don't use the row of numbers like ever. Thus, having most people being right-handed, and having all of the major controls (up/down/left/right/speed/slow/approach) within 4 inches of each other, on the right-hand side, seems to me like the way to go. But for those who have already gotten blazing fast on the row of numbers, it makes sense that adding controls waaaayyyy over on the right side of the keyboard is of little to no value, plus it uses up the arrow keys that you use occasionally and it would be annoying for them to type a '<' when that's not what the key is meant for.

To me, it seems like a question of figuring out what method is best for _new_ users, not for those of us who have been doing it for a while. Admittedly, this proposal is something I did because I am used to very similar commands to the above on the ATC simulation software from the sim lab at Kent State University, where I run the ATC lab courses. I got used to entering commands in a similar way, and can now enter hundreds of commands a minute. So I'm used to this, and you all are used to what you're used to. It comes down to which approach would be better for people with zero experience; and that's not a question I can answer. But, as a teacher, I guarantee that telling a newbie that "to make a plane go up, press up and type altitude" or "press the left arrow to turn left" is going to make more sense than directing them to "a for altitude, t for turn". For new users, it really depends on the commands' intuitiveness, speed of entry, complexity, and time to memorize them. I think my way wins on all four accounts.

@tedrek Also, you asked how I play, and its almost exclusively on the right fifth of the keyboard. To demonstrate how awesome this pull request makes entering the commands, I recorded a 45 minute run and made a video that shows the key presses, so everybody can consider the value of using the arrow keys for climbs/descents.

---

**You can view the demo video here:** [https://goo.gl/d2SOOA](https://goo.gl/d2SOOA)
First 60s is super-speed.
Best part to look it is anywhere between 10:00-20:00.

---

culebron:
Thanks for sharing the experience, I see you know it better than I do.

Actually, it's a good point to reject a command to climb to lower altitude. This shows that the controller is confusing planes or the situation. Might be worth implementing here, and I guess it's not a problem to articulate what to do in particular.

Regarding the proposal being buried, I see your point and don't mind having shortcuts, my only problem is that you're taking arrow and Ctrl+arrow keys. Is it possible to assign these actions to ALT+arrow?

tedrek:
@erikquinn, Thanks for the video it helps me understand a ton (also a new strategy or two for dealing with arrivals :wink:).

To deal with your points one by one:

> text box would not behave like a simple text box, as the arrows type stuff instead of moving the cursor

This is the only thing which prevents me from merging this, the rest of the points bear discussion but can be addressed in the longer term.

As long as there is that blinking cursor in the box and you appear to type in it the first thing people will expect from the left & right arrow keys is for it to move the cursor.

The two possible solutions I see is to implement the arrow key re-map as an option which defaults to off or to make it not appear to be a text entry box (It could actually be one, but it needs to not look or behave too much like one initially so new users are open to learning how this particular one operates).

Probably the simplest is the option and develop the non-text entry afterwards.

> seeing "^ v < >" might be less clear than "a(ltitude) t(urn)".

As I now understand your proposal, it's _all_ about the arrow keys on the keyboard not the same characters on the alphanumeric portion.  As such we don't need to use characters from elsewhere on the keyboard, instead go into unicode and use actual arrow characters.  I think the string '↻ 360 ↗ 5000 ⇛ 250' could be pretty intuitive to understand what it means and if those characters don't work checkout the [rest of the arrow block](http://www.fileformat.info/info/unicode/block/arrows/utf8test.htm).  

On the other hand, there's nothing stopping the arrow keys entering the whole words so the above could be entered with the key sequence '→360↑5000+250' but come up as 'turn right 360 climb 5000 accelerate 250' (or what have you).  This may be even more intuitive because it's clear exactly what each key press means.

Ideally I'd implement it as a shadow input box which holds the actual keys entered and a visible box which shows the expanded command, this way backspace operates per key pressed in an intuitive manner.

Taking it one step further, the canonical shortcut keys could be embedded in the shown command so there is both an explanation of the key and a reminder of the key itself.  I'm thinking something like this:
'turn right (→) 360 climb (↑) 5000 accelerate (+) 250'

> users who depend on up/down arrows recalling recent callsigns will have to instead click Ctrl+up or Ctrl+down

My suggestion would be Page Up and PageDown are available and are a very similiar concept (at least to me).  Having them as synonyms to the normal Up and Down shouldn't be a problem and are in the same portion of the keyboard so they would be usable while this mapping is in effect.

erikquinn:
@culebron Actually, I really like @tedrek's suggestion of moving callsign recalling to the Page Up/Page Down, instead of Ctrl+Up/Down or Alt+Up/Down. 

@tedrek Unicode arrow characters... Of course! Why didn't I think of that? haha that would be much better than ^<>v. I'm sure HTML can deal, by I'm not sure how much js will like playing with them; I've never tried.. I know it doesn't like Greek characters in variable names (annoying with mathy trig functions). Something to look into!

@all
If this whole deal was by default disabled, then the arrows would still behave like arrows, and campaign recall would be moved from up/down TO page up/page down. Then, that should resolve all of the advserse effects. For those like me who want it, check the box and off you go.

And though I'm really trying to lobby in support of making arrow-based commands the new standard method, I would settle with having to "turn them on" in settings instead of not having them at all. I'm willing to table the conversation of changing the primary method if that makes the proposal agreeable. (I may want to revisit it up later of course :wink:)

PS I'm away for Christmas now and left my computer at home so I won't be able to revise this thing until mid next week probably. I'm unfamiliar with pull request etiquette, so help me out here: should I close this one and make a new one when it's ready, or just toss the changes on this one once I get around to it? Also, I'm getting the sense that it's nicer to the owner for me to create a pull request on a branch I can leave alone, rather than my own master (or in this case gh-pages) where I might want to keep working, or does it not really matter? PS thanks! :)

tedrek:
Javascript shouldn't have any problems with the characters in this usage. 

For the pull requests, you should nearly always create a new branch for the PR.  If you update the branch after opening the pull request then the pull request is updated which is generally bad if the changes are unrelated.  At this point you'll probably want to close the PR and open a new one against a new branch (with a link back to this one in the description.)

As an aside on branching, I personally run a master branch which is kept in sync with zlsa/gh-pages.  This means I can muck about with the gh-pages branch in my repo as much as I want, for example sometimes I will have two or three topics going and switch the gh-pages branch between them for testing.

I'm looking forward to this change.  I can also easily see it becoming the default input method with some more work.

erikquinn:
Okay, I'll have a look in a few days. Thanks for the PR help, and thanks to both of you for the input and candor. I think we've struck a happy medium here. 

Merry Christmas!

-EQ

Maverick283:
Seeing the video @erikquinn posted, I really want a recording option that allows to play back what happened without zoom in different speeds, so you can check how you did and see if it worked out to be such an amazing flow as it was in the posted video.

As a left handed person, I believe it doesn't really matter if you are right or left handed, usage of a computer is a standard  and does not differ between the two types of people. Same as for flying a airplane, a pilot in a airbus (must) have the left hand on the stick, and the right on the thrust levers. 
I personally don't use the Num-Pad either, because I play with a mouse and thus have my right hand on the mouse most of the time. I do most of my typing with my left hand, and only add the right hand when I type a list of fixes.

For a beginner who has enough trouble getting the airplanes in order it would be a big burden to remember all the keys if they are not in any association with their actual function. 
And since we don't have a "turn left "- key on the keyboard, I think we should definitely at least preserve the function to type the actual action. 

When I started playing, the very first planes I typed the full command rather then just "d" for "direct". This is intuitive and doesn't need a lot of attention, so one can focus on the planes. As soon as this is figured out (but early enough that one is not totally used to the command way) the tutorial could maybe suggest to switch to a faster command input method, as the idea of less typing sound great to me!

PR #300, opened 2015-12-23
[Help needed] Tried to add fleet to Lufthansa and EDDM

Maverick283:
Well, I'm sorry I seem like a complete noob right now, but: It's not working :see_no_evil: 

I tried following the instructions @tedrek gives in #292, yet when testing there were almost no lufthansa planes spawned in the arrival at all, and if then only (very rarely) the e190 (which is the only plane in the "short" fleet). I can't figure it out myself, my console is not being a big help, and I'm missing the debugger I am used to from Java... 

Also I'd like to know if I can pretty much add any type of fleet, not only "long" and "short" and "default" but also "medium" or something called "midrange" or whatever as long as I keep the identifiers the same (like "dlh/midrange" in the eddm.json)? From what I saw in the code this should be possible...?

As soon as these problems clear up I will try to create more proper fleets in the airline so they can be used on the different airports!


tedrek:
@Maverick283, It all looks good to me.  The only change made to EDDM is to spawn departures from dlh/long and dlh/short and it works for me.  Arrivals spawn from dlh which is the same as dlh/default (got a b735 on the first load and an a319, a320, a321 on the second load) for arrivals.  For departures I removed everything other than the dlh/long and dlh/short lines and got two e190s and a b744 as initial departures.

My suspicion is that you either got some unlucky random numbers, or you may have run into browser caching issues.  Anytime I'm not seeing the expected behaviour I strongly suggest opening a new incognito window (If you are using Chrome) or similiar and testing from that window.  I suspect we need to implement some sort of cache buster for the .json files.

Regarding fleet names, yes you can use anything you like.  It will be a lot easier if the common ones are named the same across the board, my suggestion was to use:
- short
- medium
- long
- default

If you want to create a fleet which doesn't fit into that scheme, by all means do so we can always adjust it later if there is reason to.

Maverick283:
Okay, thanks for the incognito window hint, will try that...

Yeah, i think having short, medium, long and default as standard fleets would be good. The only reason I see to add more is if there is a unusual coverage of an airline for one airport.

Thanks for your help.

Maverick283:
ill create another pull request as soon as I got everything done...

PR #311, opened 2015-12-31
Add option for arrow-key-based user controls

erikquinn:
**The follow-up after feedback on #299.**

Adds what I called "shortKey" commands (a single string, instead of using spaces for parsing commands vs arguments. Essentially:
'turn left 250' --> '[left arrow]250'
'climb 5' --> '[up arrow]5'
etc...

**All old commands are still intact, but this method can be enabled through a game options setting ("Control Method" --> "Classic" or "Arrow Keys").**

In total, new commands include:

'[up arrow]' for climbing
'[down arrow]' for descending
'[left arrow]' for turn left
'[right arrow]' for turn right
'fh' for fly-heading
'[numpad +]' for increase speed to xxx
'[numpad -]' for decrease speed to xxx
'*' for cleared ILS XX approach

I tested it thoroughly to make sure nothing was broken, and it all seems to be in order.

Feel free to try it out: http://erikquinn.github.io/atc/


assets/scripts/aircraft.js
```diff
@@ -787,13 +858,16 @@ var Aircraft=Fiber.extend(function() {
       return ['ok', 'turn ' + direction + 'heading ' + heading_to_string(this.requested.heading), ''];
     },
     runAltitude: function(data) {
-      var split     = data.split(" ");
+      if(data[0] == "\u2B61" || data[0] == "\u2B63") {  //shortKey 'v' or '^' in use
```
tedrek, line 861: If the COMMANDS data is updated with a different key for altitude then this will break.  Would it be possible to strip the shortKey off in the parsing step and just pass the heading data?

tedrek:
Looks good, I especially like additions to the tutorial :+1:

assets/scripts/aircraft.js
erikquinn, line 861: Sure. Only reason I did it at the beginning of each heading/altitude/speed/etc's function instead of runCommand() was because of the left/right turns associated with the shortKeys for heading.... in that case, the arrow needs to be passed in order for runHeading() to know the direction of the turn. I then figured consistency was better than not, and stripped it off at the beginning of each command instead of when building the list of commands. I can definitely change it though if needed.

tedrek:
I don't see a good quick way to make the short keys work well given the distinction between left and right.  Given it's a code issue rather than gameplay I see no reason to wait for the merge.  Great work @erikquinn :+1:

PR #313, opened 2015-12-31
Support for full waypoints

tedrek:
Re-factors the multiple fix functionality out into a separate FlightManagementSystem.  This gives the ability to assign speeds and altitudes after passing a given fix.  Currently the only method of using this is to configure a set of waypoints in the aircraft arrivals (see airport-format).  This allows airports to spawn aircraft at higher altitudes and descend them as they arrive.

This is a fairly large change code wise, so I'm pretty sure there's still some bugs to it.  If a couple of people could give it a try at http://tedrek.github.io/atc and report any issues that would be greatly appreciated.


erikquinn:
Cool! This would be most useful for "climb via the SID" stuff (fancy pants airports do this,  not 100% sure how it works). having ac climb via SID and and descend via STARs would be a cool way to incorporate some of the more newly implemented procedures. Looking forward to that.

One question though: you said climb and speed after passing a fix; would it be hard to add another option for crossing restrictions (eg. "descend at pilots discretion, cross BERRA at and maintain 1-1 thousand at 2-5-0 knots")? this is common in enroute and STARs, and would be both cool and helpful to emulate.

Note: both "change altitude/hdg/speed after passing a fix" AND "maintain this alt/hdg/speed until passing a fix/line/etc" are useful. Particularly speed control on final (such as "maintain 190 knots to FINKA")

tedrek:
@erikquinn, I've tried to address the feature request in #315.

As written it supports altitude/speed until a fix only but you can have as many fixes as you want.  It is also possible as written to include a heading mode section after the final fix with a different alt/speed.  It's also possible to clear for landing automatically as part of a STAR, I don't think it's a good idea from a game play perspective but it's possible.

culebron:
Was thinking of this, great idea!

nielsvdweide:
@tedrek I was trying to test this but unable to figure out the commands. Could you give some advice?

tedrek:
@nielsvdweide, there are currently no commands exposed to the player, that will come in a later patch.  I was looking for testing to make sure nothing existing had broken with these changes as they were fairly major from a code perspective.

nielsvdweide:
@tedrek In that case, i don't experience anything different :+1:

PR #317, opened 2016-01-04
Display wind speed in compass center

erikquinn:
capture

Displays the wind speed to the user in the center of the little dot in the middle of the "compass" / wind sock thingy. Also fixes the numbers from [0, 90, 180, 270] to [360, 090, 180, 270].

### DIFF
--- a/assets/scripts/canvas.js
+++ b/assets/scripts/canvas.js
@@ -792,23 +792,34 @@ function canvas_draw_compass(cc) {
   var size    = 80;
   var size2   = size / 2;
   var padding = 16;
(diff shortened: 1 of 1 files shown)


avdg:
Hmm I wonder how it will display when there are big numbers

erikquinn:
@avdg It can handle up to two digits and still fit properly. Beyond that, I can't imagine a scenario where anything is flying in >100kt wind.... other than cows, of course. :wink: 

high spd
closeup

avdg:
Should be fine I hope :-)

PR #318, opened 2016-01-04
EDDM with SIDs and Magnetic north 2. attempt

harp71:
Finally i might have gotten Git and Github logic, many thanks to @avdg. This is (still) the reworked version of EDDM.


harp71:
@Fechulo i just found out that you already did some changes on EDDM, including SIDs. Sorry, but i had some technical issues with git and they where already there, just not online. I researched the real SIDs and STARs for EDDM. And as it is my home airfield ... would you mind if we take my version?

Fechulo:
@harp71 sure, no problem. It should be more accurate if it's your home airfield

avdg:
Hmm patch may need rework from scatch now there are conflict (changes are overwriting each other, because one edited on the same file on a different branch intended to merge in to)

tedrek:
I dealt with the conflicts here, it was just the SID block differing.  Thanks for the updates @harp71 :smile:.

PR #319, opened 2016-01-05
Add 5nm range rings (follow-up to #281) and improve airport selection dialog

erikquinn:
Follow-up to #281. This adds 5nm range rings to all airports by default. I know that's the established standard in FAA ATC, but it may not be the case in other countries... I just set them all to 5nm because that sounds better than "one-fourth of whatever the airport 'center radius' is", which is what it does if no "rr_radius" is defined. Also fills out all of the properties that may or may not have been defined in some airport.json files... eg ("has_terrain = false") for airports that don't have terrain, instead of saying nothing.

Also made some changes to clean up the airport selection dialog. See below screenshot on that. _These changes depend on #320._

Dialog may be tested at [http://erikquinn.github.io/atc/](http://erikquinn.github.io/atc/).

---

image

### DIFF
--- a/assets/airports/ebbr.json
+++ b/assets/airports/ebbr.json
@@ -1,11 +1,16 @@
 {
+  "reference": "https://www.belgocontrol.be/opersite/eaip/eAIP_Main/html/eAIP/EB-AD-2.EBBR-en-GB.html",
   "name": "Brussels-National",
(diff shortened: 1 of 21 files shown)


Maverick283:
Works like a charm for me! Looking good!

PR #322, opened 2016-01-06
Implement diverging separation rules

tedrek:
Maintain separation when aircraft are on diverging flight paths of at least 15 degrees difference and one aircraft has passed the projected course of the other.

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -4,6 +4,7 @@ zlsa.atc.Conflict = Fiber.extend(function() {
       this.aircraft = [first, second];
 
       this.distance = vlen(vsub(first.position, second.position));
+      this.distance_delta = 0;
       this.altitude = abs(first.altitude - second.altitude);
 
       this.collided = false;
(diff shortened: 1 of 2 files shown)


erikquinn:
YEEEEESSSSSS I will definitely play around with this a bit. Thanks for getting to this one!

assets/scripts/aircraft.js
```diff
@@ -166,9 +170,51 @@ zlsa.atc.Conflict = Fiber.extend(function() {
...
+         radar vectors is at least 15 degrees.
+      */
       else {
-        conflict = (this.distance < 7.4); // 4nm
-        violation = (this.distance < 5.6); // 3nm
+        var offset = abs(angle_offset(this.aircraft[0].groundTrack,
+                                      this.aircraft[1].groundTrack));
+        //TODO: Opposing course doesn't seem to be handled correctly
+        // Check for courses differing by at least 15 degrees and within 4nm
+        if ((this.distance <= 7.4) && (offset > radians(15)))
```
erikquinn: Should be **>=** 15 degrees here, seeing as 15 degree divergence _IS_ okay, while 14.9 would be a sep loss, as you noted your comment.

Other than that, everything seems to work perfectly. Badass! :smile:

tedrek:
Updated to allow precisely 15 degrees difference in heading.  Thanks for the critique @erikquinn :smile:

PR #323, opened 2016-01-07
Give 1 minute grace period for aircraft after takeoff

tedrek:
As the title says.  For 1 minute after giving the takeoff order aircraft aren't checked for proximity separation issues.

@erikquinn, should it be 1 minute from rotation or is 1 minute from the takeoff good enough?

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -75,6 +75,12 @@ zlsa.atc.Conflict = Fiber.extend(function() {
           (this.aircraft[1].altitude < 990))
         return;
 
+      // Ignore aircraft in the first minute of their flight
+      if ((game_time() - this.aircraft[0].takeoffTime < 60) ||
+          (game_time() - this.aircraft[0].takeoffTime < 60)) {
+        return;
+      }
+
       this.checkProximity();
     },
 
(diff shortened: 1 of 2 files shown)


erikquinn:
I don't feel like opening up the 65... lol but I'm fairly certain it's within 1 min after **takeoff** that they have to be established on those 15\* diverging headings. So I'd say from beginning of the takeoff roll.

Now would this apparent 60s invincibility period mean that two aircraft departing simultaneously toward the intersection of their two runways (illegal) would NOT trigger a CA? Or is that logic separate from the "proximity separation" you mentioned?

tedrek:
Excellent, it's coded as 1 minute from the beginning of the takeoff roll, I figured for our purposes it probably wasn't going to matter too much anyway.

The 60s period won't give invincibility for intersecting runway issues.  My thinking is that it only applies to airborne aircraft.  The intersecting and same runway separation issues need to apply regardless of the aircraft being airborne, however it remains to be written.

As it is right now, there are no conflict alerts of any kind below 990 feet MSL.  My intention is to drop it down to 200 feet AGL as a rough radar horizon effect and then let this grace period and the diverging heading patch handle airborne stuff.

Issue #330, opened 2016-01-08
Getting a lock on ILS


montypython112:
I just cannot seem to get any arrivals to actually land, despite having them at a low enough altitude and decent heading they seem to simply fly right over the runway. The tutorial is of no help as it mentions using asterisks, which as far as I can tell do not work, so I have been using the land command. 
<img width="457" alt="screen shot 2016-01-08 at 12 32 12 pm" src="https://cloud.githubusercontent.com/assets/16603501/12205601/bd38dee8-b609-11e5-86a2-b83e6374e2fb.png">
AAL600 should by all accounts have locked in at some point right?

Issue #262, opened 2015-11-23
Another Reality thing: Approach clearances 


Maverick283:
In this game, whenever a plane gets close to the localizer it is being cleared to land (BAW001 L 28L).

In reality, a plane is cleared to fly the approach, and then later cleared to land.
"Speedbird 001, [distance]nm out of [initial approach fix], fly [heading] until you intercept the localizer for runway [runway], cleared for the [ILS/VOR/RNAV etc.] Approach [runway]". 
Then the landing clearance would be:
"Speedbird 001, wind [degrees],[knots], [runway] cleared to land." 

Both of those are off the top of my head and might need some double checking...

What I'm on about though is that we simply drop one (rather important) clearance and should think about introducing it.

culebron:
I think this is a work for two ATC, approach and tower. If you play this alone, it's just too big a workload.

On landings, I would think of planes themselves going around when there is another plane on the runway.

nielsvdweide:
@Maverick283 You are right about clearing the approach as well as the clearance to land, but as a one man ATC operation it is just too much information and adds a too high workload. Clearance for the approach is normally given by the Approach controller and when the airplane is on it's final track to land they will be switched to Tower. They will then give them the landing clearance! 

No point in adding one of them. I agree it will be more realistic but thing will just get too complicated

Maverick283:
Okay, we got this covered now I guess xD.

I agree that it increases workload, and it is probably not worth the added reality...

Can I close this issue myself and just don't know how or does @zlsa have to do this?

glangford:
You should be able to close any issue that you opened; look at the bottom here for a button called 'close issue' (or something like that):

https://github.com/zlsa/atc/issues/262

Issue #232, opened 2015-11-11
Multiple language support


Maverick283:
Another idea for an enhancement:

Based on Google Analytics or other means, users could get the explanations and other in game information in their own language. The ATC would obviously stay english. 

That way more users might start playing.

This again can wait, and just adds to the collection of ideas for later implementation.

MaicomMR:
I can help to make a Braziliam Portuguease(PT-BR) translation, call me at need :smile:  :+1:

Maverick283:
I'd be down for German and Spanish...

erikquinn:
The ATC repository is being migrated to it's new home at [https://github.com/openscope/openscope](https://github.com/openscope/openscope), 
and thus, all issues are being closed. If this is still an issue with the latest version of the sim 
(accessible at [http://www.openscope.co](http://www.openscope.co)), or is a feature you still think we are lacking, 
please reopen the issue at the new repo.

*Please note that the vast majority of these issues have been copied to the new repository, or else are covered by other issues created there. See the below screenshot for what it looks like when your issue is known in the new repo:*

![image](https://cloud.githubusercontent.com/assets/5103735/22630816/5a32d794-ebce-11e6-9c46-d61a262e2b23.png)


 Thank you!


Closing this issue.

PR #331, opened 2016-01-09
"*" Not working as a landing command, changed text accordingly

Maverick283:
Hope I got this all right... Issue #330 brought to attention that the "*" command is not a valid command for the ILS approach. I changed the tutorial to show both the "Land" command and its abbreviation, "l" to use for getting an aircraft locked on the ILS.

### DIFF
--- a/assets/scripts/tutorial.js
+++ b/assets/scripts/tutorial.js
@@ -261,7 +261,7 @@ function tutorial_init_pre() {
 
   tutorial_step({
     title:    "Approach Clearances, part 1",
-    text:     ["You can clear aircraft for an ILS approach with the "ILS" command, followed by a runway name. Before you can do so, however,",
+    text:     ["You can clear aircraft for an ILS approach with the &quot;ILS&quot; command, followed by a runway name. Before you can do so, however,",
                "it must be on a heading that will cross the runway's extended centerline, that is no more than 30 degrees offset from the",
                "runway's heading. Once we eventually give them an approach clearance, you can expect aircraft to capture the ILS's localizer",
                "once they're within a few degrees of the extended centerline."
--- a/assets/scripts/tutorial.js
+++ b/assets/scripts/tutorial.js
@@ -261,7 +261,7 @@ function tutorial_init_pre() {
 
   tutorial_step({
     title:    "Approach Clearances, part 1",
-    text:     ["You can clear aircraft for an ILS approach with the "Land" command, followed by a runway name. Before you can do so, however,",
+    text:     ["You can clear aircraft for an ILS approach with the "ILS" command, followed by a runway name. Before you can do so, however,",
                "it must be on a heading that will cross the runway's extended centerline, that is no more than 30 degrees offset from the",
                "runway's heading. Once we eventually give them an approach clearance, you can expect aircraft to capture the ILS's localizer",
                "once they're within a few degrees of the extended centerline."
@@ -275,7 +275,7 @@ function tutorial_init_pre() {
 
   tutorial_step({
     title:    "Approach Clearances, part 2",
-    text:     ["When you have the aircraft facing the right direction, just select it and type &lsquo;l &lt;runway&gt;&rsquo;",
+    text:     ["When you have the aircraft facing the right direction, just select it and type &lsquo;i &lt;runway&gt;&rsquo;",
                "with the runway that&rsquo;s in front of it. Once it's close enough to capture the localizer, the assigned altitude on its strip",
                "will change to &lsquo;ILS locked&rsquo; (meaning the aircraft is capable of guiding itself down to the runway via",
                "the Instrument Landing System), and the assigned heading should now show the runway to which it has an approach clearance."
--- a/assets/scripts/tutorial.js
+++ b/assets/scripts/tutorial.js
@@ -275,7 +275,7 @@ function tutorial_init_pre() {
 
   tutorial_step({
     title:    "Approach Clearances, part 2",
-    text:     ["When you have the aircraft facing the right direction, just select it and type &lsquo;l&lt;runway&gt;&rsquo;",
+    text:     ["When you have the aircraft facing the right direction, just select it and type &lsquo;l &lt;runway&gt;&rsquo;",
                "with the runway that&rsquo;s in front of it. Once it's close enough to capture the localizer, the assigned altitude on its strip",
                "will change to &lsquo;ILS locked&rsquo; (meaning the aircraft is capable of guiding itself down to the runway via",
                "the Instrument Landing System), and the assigned heading should now show the runway to which it has an approach clearance."
--- a/assets/scripts/tutorial.js
+++ b/assets/scripts/tutorial.js
@@ -261,7 +261,7 @@ function tutorial_init_pre() {
 
   tutorial_step({
     title:    "Approach Clearances, part 1",
-    text:     ["You can clear aircraft for an ILS approach with the asterisk (*) key, followed by a runway name. Before you can do so, however,",
+    text:     ["You can clear aircraft for an ILS approach with the "Land" command, followed by a runway name. Before you can do so, however,",
                "it must be on a heading that will cross the runway's extended centerline, that is no more than 30 degrees offset from the",
                "runway's heading. Once we eventually give them an approach clearance, you can expect aircraft to capture the ILS's localizer",
                "once they're within a few degrees of the extended centerline."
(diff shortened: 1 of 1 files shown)


erikquinn:
The asterisk **key** works, but doing shift+8 doesn't, because the asterisk key actually enters a more appealing star (Unicode 2B50) instead of a normal asterisk symbol.

To make shift+8 also work the same way, copy lines 234-238 from input.js and add another case "56" and an "if(e.shiftKey)" before the rest of the stuff. Or, just remove the case that creates the star, but that will also require changes to aircraft.js as well.

However, I'm pretty sure that "l28r" as amended will not work, as runLanding() probably splices by spaces. Unless that's not true, it should be "l 28r" instead of "l28r".

Sidenote: If we wanted to add 'i' and 'ils' to the runLanding synonyms, then it would be okay to have a tutorial direct them to use 'i 28r', but honestly we are clearing them for the ils approach, NOT clearing them to land, which is a distinction that has always irked me somewhat, so in my opinion, because of that issue, I'd really recommend against telling people to use 'land' at all, since we aren't clearing them to land, but rather clearing them for an approach. My thoughts anyway.

Maverick283:
Yeah, you are definitely right, I added the space.

I think we should keep the tutorial to show the "land" command (or rather the "l" command) because of the problem you addressed. My asterisk (on a QWERTZ keyboard) is on the same button as the + Symbol. To use it, [Shift] + "+" needs to be pressed which results in zooming in and no input to the text field. Thus I tried copying a asterisk from somewhere else, and past it in the text box. I tried both, "\* 28L" and "*28L", but neither command was accepted.

Maybe using "keys" is the wrong approach on for this? Why not use symbols instead?  

As for the landing clearances... I brought this issue to attention in #262 but we came to the conclusion to leave it the way it is it would increase the workload for the controller a lot!

erikquinn:
I reviewed #262, and while I agree with everything you were saying in it, I also have to also agree with what everyone else said there: requiring separate approach clearance and then landing clearance commands adds more workload than it's worth. I'm all for realism, but as far as using the sim for practice, making me enter both commands doesn't add value.

What I'm suggesting is that we continue with a single command, but we add 'ils' as a synonym for runLanding(), so from the user's perspective, they don't clear a/c to "land", but instead clear it for "ils". And the tutorial should guide people to then use 'i 28r' instead of 'l 28r' (as an alternative to the \* key). Since we are descending and vectoring our arrivals in an attempt to get them onto the localizer, it makes much more sense to present it this way, and ignore the need to explicitly give a landing clearance (as opposed to giving a landing clearance and ignoring the need to explicitly give an approach clearance). Based on what you were saying in #262, this should probably sound like an attractive resolution to you too I would think.

One more note: Now that I think of it, laptops almost never have a numpad, so its crazy to expect everybody to use the asterisk by the numpad. Thus, we definitely need to make shift+8 have the same behavior. I could get to that tomorrow at the earliest though as I am out of town for a funeral.

Maverick283:
Okay, I'll look into that code later and see if I can add/change the "l" to an "i".

As for the Shift+8 thing: If it type Shift + 8 I get a "(", not an Asterisks, as the symbols are layouted different on a QWERTZ Keyboard, and I'm assuming French Keyboards are different then QWERTY and QWERTZ and so on. So what I'm suggesting is that either instead of the actual key to be detected (Shift + 8) the program detects the symbol that is put (Asterisk) when it is pressed. 
That way, when I do Shift + "+", I get the same behavior you get when pressing Shift + 8. 
Or we create a smart tutorial, that knows what symbol is typed when pressing Shift + 8 on the keyboard the user is using, and displays it, for example so the text would go: "In order for a plane to align with the ILS, you have to type the command "(28L" " for me and "In order for a plane to align with the ILS, you have to type the command "*28L" " for you...

erikquinn:
Alright, I'm with ya now. in that case, we will have to remove the fancy Unicode star, which is totally fine because its only cosmetic. I honestly just didn't know there were such differences in keyboard layouts.

I made the changes so I know where all this stuff is already; I can just remove the fancy star and add 'i'/'ils'. It'll be a super quick fix, and then I'll post a link here to to see if its working for you and montypython

erikquinn:
I made a new branch "fix-asterisk" and merged them into my gh-pages.

@Maverick283 @montypython112 When able, could you have a look at [http://erikquinn.github.io/atc/](http://erikquinn.github.io/atc/) and verify the "*28r" command works on your keyboard layout now?

Maverick283:
@erikquinn, it works if I use my external keyboard, that has a numpad, yet when I use my laptop, it will still zoom in rather then writing a Asterisk. Because the + is on the same key. Pressing Shift + "+" results in zooming in, because it is the key that is detected.

Maverick283:
Did you also implement the "i 28r" to work as an addition to the "l 28r"? If so, I'll rechange the tutorial and finally contribute some (very little) code to this whole thing rather then just my usual blablabla xD

erikquinn:
@Maverick283 yeah, I added 'i 28r' up there. Now that I think about it though, it probably would be nice to get rid of the darn space and make it 'i28r'... regardless, there are bigger problems.
image
Assuming this is the keyboard layout you use, I see what you're talking about, and I _don't_ see an easy way to make it play nice with non-qwerty layouts. I thought maybe it could be fixed by stopping all use of [`e.which`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which), and instead using [`e.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key), which references the _content_ of the key (ie '*') instead of the key code (ie '106'), because the latter is different on different layouts, BUT it turns out Chrome doesn't like `e.key` anyway.

Apparently, according to [the internet](http://stackoverflow.com/a/12825956/5774767), the inability to deal well with multiple keyboard layouts with key events is a pretty common problem that doesn't have an attractive solution... which I find really surprising.

Note: the reason it doesn't work on the German QWERTZ layout (pictured above), is that the `* / + / ~` key (left of enter) has a code 187 returned by e.which. While the US QWERTY layout, the key in that location shows `} / ]` and returns code 221, and the `+ / =` key (left of backspace) returns code 187. Thus, since the code is set to reset zoom on [code 187] and zoom in on [shift + code 187], it is actually on the german qwertz layout _resetting the zoom_ when the key labeled "+" is pressed, and zooming in when shift + "+" (aka the \* symbol) is pressed. Thus, the attempt to make the asterisk from the keypad (not numpad) work as well is impossible, since it is placed elsewhere in different layouts.

If only there was a way to reference the CHARACTER being pressed instead of dealing with key codes... this is getting beyond my expertise though, for sure. I wonder if any of the "big dogs" around here have advice on this?

Maverick283:
I'm like a little chihuahua... But what if we don't use the Asterisk but another character that would work instead? For example the "@" symbol? (Think it is not used yet). As far as I know, it is [Shift] + [2] for QWERTY, and it is [Alt Gr] + [q] for QWERTZ. For QWERTY I think it is great since it is right next to the 1 & 3, so the numbers used most as the first digit when assigning a runway. Also it makes sense to have a aircraft land "at" a runway ;)

If the [Alt Gr] gives you trouble (I know this is unique to QWERTZ Layouts), then maybe the "@" isn't the best solution. Then "#" comes to mind, and so on...

erikquinn:
@ would still have the same problem. Though we can both enter an "@" symbol, the code only cares about the "keyboard event", which will tell you the "code" of the key pressed.
- on qwerty: `[shift] + [2] --> code 50`
- on qwertz: `[alt gr] + [q] --> code 81`

Since we differentiate by the key code in input.js (line 171), the two @ entries are not the same. We could add both, but then my "q" key would enter an "@" symbol... it just gets messy. That's why we would want to be able to differentiate not by key codes but by the character that the key would type, but there doesn't seem to be a way to do so. We just haven't run into this problem yet because up to this point, we never used many designated keys.

Of course, this is all only a problem because I'm trying to eliminate the need to hit the spacebar (as it is way out of the way when using arrow-key controls, which I still strongly believe are the way to go). The idea is that you could type `fh250` and then strike the "*" key (wherever it may be), and the preceding space would be entered for you. Thus, to type `fh250 *28r`, you go [f] [h] [2] [5] [0] [*] [2] [8] [r] [enter], and the command goes in properly. If there's no good solution to this problem, maybe I would just have to deal with entering the space manually. It just seems insane that we can't get it to work this way.

And not at all! Keep bringing on the ideas; I want to see this problem solved just as much as the next guy :smile:

Maverick283:
Okay, so we are basically looking for a key that has no usage at all at the moment...? What about "/"? It is on the numpad, a key that I don't believe is yet used on QWERTY, and would be Shift + 7 for QWERTZ. As I don't want to be egoistic, I started looking where the "/" is on other keyboards (AZERTY has it where  the QWERTY ">" is), and it gets worse the more keyboards there are to add...

Two solutions come to mind:
- Let the user select his layout and execute commands based on his selection
- Find a way to detect not the key but the symbol pressed

For the user selection, that could go hand in hand with a translation as suggested in #232.

And I will now create a new issue as this is getting out of hand here...

assets/scripts/tutorial.js
```diff
@@ -292,7 +292,7 @@ function tutorial_init_pre() {
     text:     ["You may choose to enter one command at a time, but air traffic controllers usually do multiple. Particularly in approach clearances,",
                "they follow an acronym &ldquo;PTAC&rdquo; for the four elements of an approach clearance, the &lsquo;T&rsquo; and &lsquo;C&rsquo; of which",
                "stand for &lsquo;Turn&rsquo; and &lsquo;Clearance&rsquo;, both of which we entered separately in this tutorial. Though longer, it is both ",
-               "easier and more real-world accurate to enter them together, like this: &lsquo;fh250 *28r&rsquo;."
+               "easier and more real-world accurate to enter them together, like this: &lsquo;fh250 l28r&rsquo;."
```
tedrek, line 295: I believe this should include a space like `l 28r`

Issue #326, opened 2016-01-07
[Enhancement] Considerations for Wake Turbulence


erikquinn:
At some point, we should think of taking wake turbulence effects into account. There are some (albeit rather complicated) rules that govern the way ATC is expected to provide additional separation due to the effects of wake turbulence. And they're getting more complicated with this whole new thing about "recat", or Wake Turbulence Recategorization, where the wake turbulence separation minima are decreased to what is actually needed instead of having a blanket figure that is safe in some situations and excessive in others.

Anyway, this would be a great thing that advanced users could choose to turn on via the settings menu that would really take the simulation to a place where it's a legitimately excellent training tool for aspiring (or hell, even _current_) radar controllers.

---

Items to prep before work can begin adding separation logic:
- [x] **Update all aircraft json files to have more information**
  - [x] **SRS Category**
  - [x] **Wake Turbulence Classification**
  - [x] _Wake Turb Recat Category (optional)_
  - [x] _LAHSO Category (optional)_
  - [x] _Powerplant type & number (optional)_
- [ ] **Somehow store matrix of separation minima**
  - [ ] **En-trail radar separation: "following"**
  - [ ] **Subsequent approaches: "landing behind"**
  - [ ] **Time/Distance based sep for same-rwy departures**
  - [ ] **Time-based sep for departures when H/S departs intersecting rwy**

---

Things to remember when implementing the wake turbulence separation logic:
- Applicable separation minima is always different based on the aircraft and the situation
- Trailing box (2,500' laterally, both directions, x miles behind, to but not including 1,000' below)

erikquinn:
The ATC repository is being migrated to it's new home at [https://github.com/openscope/openscope](https://github.com/openscope/openscope), 
and thus, all issues are being closed. If this is still an issue with the latest version of the sim 
(accessible at [http://www.openscope.co](http://www.openscope.co)), or is a feature you still think we are lacking, 
please reopen the issue at the new repo.

*Please note that the vast majority of these issues have been copied to the new repository, or else are covered by other issues created there. See the below screenshot for what it looks like when your issue is known in the new repo:*

![image](https://cloud.githubusercontent.com/assets/5103735/22630816/5a32d794-ebce-11e6-9c46-d61a262e2b23.png)


 Thank you!


Closing this issue.

Issue #298, opened 2015-12-22
New json-structure for aircrafts based on performance database


harp71:
I started a new issue from zlsa/atc/issues/220 (Some thoughts on enhancements)

As to the reorganisation of the aircrafts performance data i did research as good as i could, and i found some interessting data. Regarding the rate of realism however i had to conlude, that one cannot simulate reality a 100%, as there are so many variables in play. An A320 for example can fly short to mid-range (fuel), with 20 or 150 passengers and their luggage. This would result in a totaly different performance for the same aircraft.
On the other hand the simulation starts just above 10000 feet and for ATC-purposes it doesn´t realy mater, if there is a ton or 10 more weight on an aircraft.

I´d like to make a proposal as to the new json-structure of aircrafts, based on the relevant data for this simulation and the database (eurocontrol, see below 1.) on aircraft performances i found, as follows. I suggest to rewrite all existing aircraft (i offered to do that) and leave some kind of reference (_comment) in each file for new aircrafts, where to get the data from. 

Here my example for an A320, reference here: https://contentzone.eurocontrol.int/aircraftperformance/details.aspx?ICAO=A320&

``` javascript
{
 "_comment": "get aircraft data from: https://contentzone.eurocontrol.int/aircraftperformance/",
 "manufacturer": "Airbus",
 "name": "Airbus A320",
 "type": {
     "icao": "A320", 
     "typecode": "L2J", // = Landplane, 2-Engines, Jet-Engine (see below 2.)
 },
 "takeoff": {
     "iasv2": 145, // minimum takeoff speed
     "runway": 2190, // minimum runway length
     "wtc": "M" // Wake Turbulence Category (L/M/H/SH, see below 4.)
 },
 "climb": {
     "ias50": 175, // Indicated Air Speed takeoff to 5000 feet
     "ias150": 290, // Indicated Air Speed 5000 feet to 15000 feet
     "roc50": 2500, // Rate Of Climb takeoff to 5000 feet
     "roc150": 2000 // Rate Of Climb 5000 feet to 15000 feet
 },
 "approach": {
     "ias100": 290, // Indicated Air Speed above 10000 feet
     "iasapp": 250, // Indicated Air Speed below 10000 feet
     "mcs": 210, // Minimum control speeds below 10000 feet (see below 5.)
     "rod100": 3500, // Rate Of Descent above 10000 feet
     "rodapp": 1000 // Rate Of Descent below 10000 feet
 },
 "landing": {
[...]

PR #332, opened 2016-01-10
Prep for #326: Add more aircraft categorizations/characteristics

erikquinn:
All changes are back-end only, and shouldn't affect anything. Just provides more information that will be referenced later on for various separation & wake turbulence rules. Large number of lines changed, but really it's just the same thing being added to all 64 aircraft, with different values of course. For example:

```
"engines": {
    "number": 2,
    "type": "J"
  },
  "weightclass": "H",
  "category": {
    "srs": 3,
    "lahso": 9,
    "recat": "B"
  },
```

Helps prep for work on #326, and would check boxes 1a-1e.

### DIFF
--- a/assets/aircraft/a124.json
+++ b/assets/aircraft/a124.json
@@ -1,9 +1,17 @@
 {
   "name": "Antonov 124 Ruslan",
   "icao": "A124",
-  "wake": "heavy",
+  "engines": {
+    "number": 4,
+    "type": "J"
+  },
+  "weightclass": "H",
+  "category": {
+    "srs": 3,
+    "lahso": null,
+    "recat": "B"
+  },
   "rate": {
-    "turn":       0.7,
     "ascent":     1000,
     "descent":    2000,
     "accelerate": 4,
--- a/assets/aircraft/a306.json
+++ b/assets/aircraft/a306.json
@@ -1,9 +1,17 @@
 {
   "name": "Airbus A300-600",
   "icao": "A306",
-  "wake": "medium",
+  "engines": {
+    "number": 2,
+    "type": "J"
+  },
+  "weightclass": "H",
+  "category": {
+    "srs": 3,
+    "lahso": 7,
+    "recat": "C"
+  },
   "rate": {
-    "turn":       1.1,
     "ascent":     1800,
     "descent":    2100,
     "accelerate": 8,
--- a/assets/aircraft/a318.json
+++ b/assets/aircraft/a318.json
@@ -1,9 +1,17 @@
 {
   "name": "Airbus A318",
   "icao": "A318",
-  "wake": "medium",
+  "engines": {
+    "number": 2,
+    "type": "J"
+  },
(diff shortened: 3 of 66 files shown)


harp71:
@erikquinn, where are the values from?

erikquinn:
@harp71 pain-in-the-butt research honestly. mostly from appendix A in the 65 (for powerplant num/type, weight class, srs cat & lahso group), and the wake turb recat categories I had to determine based on rules from [FAA JO 7110.659](https://www.faa.gov/regulations_policies/orders_notices/index.cfm/go/document.information/documentID/1026926) and MTOW and wingspan values for each a/c type that I got from wikipedia.

ps neither source used has any more useful information related to performance, so the vast majority of the performance data you were talking about adding in #298 isn't data I found a magical database of, if that's what you were hoping for. haha that would still be a separate (and very worthwhile) effort.

harp71:
@erikquinn what is "appendix A in the 65"?

erikquinn:
@harp71 sorry, I mean as in [FAA JO 7110.65](http://www.faa.gov/regulations_policies/orders_notices/index.cfm/go/document.current/documentNumber/7110.65), the singular document that lays out all the rules for FAA air traffic control. In there at the end is Appendix A, where you can look up info about aircraft (but the breadth of information for each aircraft is only the bare minimum needed for controller reference).

harp71:
@erikquinn unfortunately i have no federal email or PIV card. Can you get me the PDF or would there be men in black suits on your doorstep, if you do?

erikquinn:
@harp71 I don't have any federal email nor a piv either... it should just be regularly downloadable at [http://www.faa.gov/documentlibrary/media/order/atc.pdf](http://www.faa.gov/documentlibrary/media/order/atc.pdf). If that is prompting you to login or something (its doing nothing of the sort to me), maybe it is only available for download in the US? If so, I can ask about its distribution, and will get you a copy only if its allowable, as I am in the process of trying to become a controller with the FAA (wouldn't want to jeopardize that).

Maverick283:
It is definitely downloadable from other countries then the US and it comes up after a simple [google search](https://www.google.de/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=FAA+air+traffic+control+pdf).

tedrek:
@erikquinn, any chance you could update dh8d.json as well?  It was added since you opened the pull request.

I'm also wondering if these values should be made mandatory, or provided with a default if unspecified in the .json.

erikquinn:
@tedrek done.

And yes, I would say they'd have to be mandatory, else throw some sort of critical error when it tries to load so future aircraft definitely have those properties (otherwise there would be no way of knowing how to apply separation rules). I'd avoid a default; treating undefined a/c like a B737 (for ex) would work, but really we should be using real info instead, as it shouldn't be hard to find.

I can address that later on when aircraft.js actually starts to parse these properties into the aircraft object.

PR #344, opened 2016-01-17
Make all altitude entries in hundreds of feet (aka flight levels) - resolves #342

### LINKED ISSUE #342: c 9 means climb to 9k yet c 10 means climb to 1k.
I could have sworn that I used c 10 to direct departing craft to 10k. Now it only gets them to 1k. If c 9 means 9k why wouldn't c 10 send them to 10k? 

I would debate that 10 as 10k would be more useful than 10 as 1k. If I ever need a craft to descend to 1k I'd be happy to type out 1000. Why the change? Or is this as it should be?

erikquinn:
Resolves the issues brought up in and the resulting discussions from #342.

The tutorial and wiki were updated to show that you must now enter altitudes in hundreds of feet, eg:
- 3,000ft = "30"
- 7,500 = "75"
- FL180 (18,000) = "180"

### DIFF
--- a/assets/scripts/util.js
+++ b/assets/scripts/util.js
@@ -276,11 +276,15 @@ var radio_names = {
   ".":"point",
 };
 
-var radio_compass_names = {
+var radio_cardinalDir_names = {
   "n":"north",
+  "nw":"northwest",
   "w":"west",
+  "sw":"southwest",
   "s":"south",
+  "se":"southeast",
   "e":"east",
+  "ne":"northeast"
 };
 
 var radio_runway_names = clone(radio_names);
@@ -289,7 +293,7 @@ radio_runway_names.l = "left";
 radio_runway_names.c = "center";
 radio_runway_names.r = "right";
 
-function radio(input) {
+function radio_spellOut(input) {
   input = input + "";
   input = input.toLowerCase();
   var s = [];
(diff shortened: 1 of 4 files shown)


assets/scripts/util.js
```diff
@@ -332,14 +336,14 @@ function radio_trend(category, measured, target) {
   return CATEGORIES[category][2];
 }
 
-var DIRECTIONS = ["N", "NE", "E", "SE", "S", "SW", "W", "NW"];
+function radio_altitude(alt) {
+  if(alt >= 18000) return "flight level " + parseInt(alt.toString().substr(0,3));
```
avdg: Even when fl 1000 might not be realistic I wouldn't advise to just take the first 3 digits (I prefer division to get that number). Also what about altitudes not rounded by 100?

erikquinn: Yeah you may be right. And now that I think about it, in Europe they can have some flight levels below 10,000ft, eg 7000 being "Flight Level Seven Zero", which would be a different number of digits... Just for the future it should probably be set up as you suggested.

And I'm not sure what you mean about altitudes that aren't rounded by 100... But all that ATC would see is precision to the hundreds-of-ft level, if that's what you're getting at. Expressing tens of feet is not needed either in communications or in display, so I am not worried about that.

harp71:
go FL ;)

PR #349, opened 2016-01-20
Change a/c warning/violation ring appearance

### LINKED ISSUE #346: Warning & Separation Circles misleading
The white warning circles and red separation circles are misleading... They have a radius of 2.5nm, and the separation required is 3.0nm, resulting in a/c generating sep losses when they are still outside of the circle. 

I recommend that we:
- Change the radius from 2.5nm to 3nm
- dim the "warning" from the bright white to barely visible white

Example:
image

erikquinn:
Resolves #346.
- added new km(nm) and nm(km) fxns to convert kilometers-nautical miles
- increased ring size from 2.5nm radius to 3nm radius
- dimmed warning ring's color (still white, just faint)

Note: the majority of the line changes in the code are just related to the addition of the km() and nm() functions, and renaming an old (and often used) function from 'km()' to 'km_to_px()'.

---

Old:
image

---

Proposed:
image

---

Use the 5nm range rings around the airport in the background to see the difference in scale; also note that the violation shown in the top image is due to the rings being 2.5nm radius and separation minima being 3.0nm radius-- this is resolved in the proposed lower image.

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -171,8 +171,8 @@ zlsa.atc.Conflict = Fiber.extend(function() {
         }
         else
         {
-          conflict = (this.distance < 5.2); // 2.8nm
-          violation = (this.distance < 4.6); // 2.5nm
+          conflict = (this.distance < km(2.8));   // 2.8nm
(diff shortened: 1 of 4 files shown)


tedrek:
:+1:

PR #356, opened 2016-01-22
Dynamic climb rate

### LINKED ISSUE #339: Acceleration and climb are way off
I just made a comparison between a real flight, A320 leaving EDDM via SID MAMOR and the ATC-Simulation. I did that, because i always found the climb-rate to be way of reality. I also found the acceleration was way to fast. I don´t know, if acceleration and climb-rate are calculated together, but they should be ... after all, climb rate affects speed-gain and vis versa. 

Long story short, here are two screenshots. 

image

image

I should mention, that i set the max speed of the sim-A320 to 400 and gradually set higher speeds as he goes, because otherwise, the sim-A320 will be at 400 knots with 1900 feet height. I also should mention, that of course this is the same SID, same aircraft, same runway, same departure heading and of course same location of to planes where the screenshot is taken.

The real data are taken from a Lufthansa A320 from Munich to Berlin using flightradar24.com:
http://www.flightradar24.com/data/airplanes/d-aipa/#8844290

erikquinn:
Adds new values and new logic to make climb and descent performance very believable. Updated the base `rate.climb` and `rate.descent` values to more reasonable figures, and also applied a trend based on air density, since (to my knowledge), without geting super-crazy-accurate with unnecessary math, the most simple, yet accurate way to approximate climb performance is by relating it to density altitude. See the graph below:

image

The climb rate is decreased proportionally to density altitude, and ends up at +500fpm (jets) or +100fpm (turboprop/piston) at the aircraft's model's service ceiling, aka `aircraft.model.ceiling`.

_Partially_ resolves #339 (fixed climb rate, but acceleration isn't changed).

Feel free to give it a whirl: [http://erikquinn.github.io/atc](http://erikquinn.github.io/atc)

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -252,7 +252,7 @@ var Model=Fiber.extend(function() {
 
       this.rate = {
         turn:       0, // radians per second
-        ascent:     0, // feet per second
+        climb:      0, // feet per second
         descent:    0,
         accelerate: 0, // knots per second
         decelerate: 0,
@@ -285,8 +285,8 @@ var Model=Fiber.extend(function() {
       if(data.ceiling) this.ceiling = data.ceiling;
       if(data.rate) {
         this.rate         = data.rate;
-        this.rate.ascent  = this.rate.ascent  / 60;
-        this.rate.descent = this.rate.descent / 60;
+        this.rate.climb  = this.rate.climb;
+        this.rate.descent = this.rate.descent;
       }
 
       if(data.runway) this.runway = data.runway;
@@ -1805,14 +1805,32 @@ var Aircraft=Fiber.extend(function() {
         // ALTITUDE
 
         var distance = null;
-        var expedite_factor = 1.7;
+        var expedite_factor = 1.5;
         this.trend = 0;
         if(this.target.altitude < this.altitude - 0.02) {
-          distance = -this.model.rate.descent * game_delta() / expedite_factor;
+          distance = -this.model.rate.descent/60 * game_delta();
           if(this.mode == "landing") distance *= 3;
           this.trend -= 1;
         } else if(this.target.altitude > this.altitude + 0.02) {
-          distance =  this.model.rate.ascent  * game_delta() / expedite_factor;
+          function getClimbRate(ac) {
+            var a = ac.altitude;
+            var r = ac.model.rate.climb;
+            var c = ac.model.ceiling;
+            if(ac.model.engines.type == "J") var serviceCeilingClimbRate = 500;
+            else var serviceCeilingClimbRate = 100;
+            if(ac.altitude < 36152) { // in troposphere
+              var cr_uncorr = r*420.7* ((1.232*Math.pow((518.6 - 0.00356*a)/518.6, 5.256)) / (518.6 - 0.00356*a));
+              var cr_current = cr_uncorr - (a/c*cr_uncorr) + (a/c*serviceCeilingClimbRate);
+            }
+            else { // in lower stratosphere
+              //re-do for lower stratosphere
+              //Reference: https://www.grc.nasa.gov/www/k-12/rocket/atmos.html 
+              //also recommend using graphing calc from desmos.com
+            }
+            return cr_current;
+          }
+          var climbrate = getClimbRate(this);
+          distance = climbrate/60 * game_delta();
           if(this.mode == "landing") distance *= 1.5;
           this.trend = 1;
         }
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -784,6 +784,24 @@ var Aircraft=Fiber.extend(function() {
       }
       return airline_get(this.airline).callsign + " " + radio_spellOut(callsign) + heavy;
     },
+    getClimbRate: function() {
+      var a = this.altitude;
+      var r = this.model.rate.climb;
+      var c = this.model.ceiling;
+      if(this.model.engines.type == "J") var serviceCeilingClimbRate = 500;
+      else var serviceCeilingClimbRate = 100;
+      if(this.altitude < 36152) { // in troposphere
+        var cr_uncorr = r*420.7* ((1.232*Math.pow((518.6 - 0.00356*a)/518.6, 5.256)) / (518.6 - 0.00356*a));
+        var cr_current = cr_uncorr - (a/c*cr_uncorr) + (a/c*serviceCeilingClimbRate);
+      }
+      else { // in lower stratosphere
+        //re-do for lower stratosphere
+        //Reference: https://www.grc.nasa.gov/www/k-12/rocket/atmos.html 
+        //also recommend using graphing calc from desmos.com
+        return this.model.rate.climb; // <-- NOT VALID! Just a placeholder!
+      }
+      return cr_current;
+    },
     hideStrip: function() {
       this.html.hide(600);
     },
@@ -1812,24 +1830,7 @@ var Aircraft=Fiber.extend(function() {
           if(this.mode == "landing") distance *= 3;
           this.trend -= 1;
         } else if(this.target.altitude > this.altitude + 0.02) {
-          function getClimbRate(ac) {
-            var a = ac.altitude;
-            var r = ac.model.rate.climb;
-            var c = ac.model.ceiling;
-            if(ac.model.engines.type == "J") var serviceCeilingClimbRate = 500;
-            else var serviceCeilingClimbRate = 100;
-            if(ac.altitude < 36152) { // in troposphere
-              var cr_uncorr = r*420.7* ((1.232*Math.pow((518.6 - 0.00356*a)/518.6, 5.256)) / (518.6 - 0.00356*a));
-              var cr_current = cr_uncorr - (a/c*cr_uncorr) + (a/c*serviceCeilingClimbRate);
-            }
-            else { // in lower stratosphere
-              //re-do for lower stratosphere
-              //Reference: https://www.grc.nasa.gov/www/k-12/rocket/atmos.html 
-              //also recommend using graphing calc from desmos.com
-            }
-            return cr_current;
-          }
-          var climbrate = getClimbRate(this);
+          var climbrate = this.getClimbRate();
           distance = climbrate/60 * game_delta();
           if(this.mode == "landing") distance *= 1.5;
           this.trend = 1;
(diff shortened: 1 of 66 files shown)


assets/scripts/aircraft.js
```diff
@@ -1805,14 +1805,32 @@ var Aircraft=Fiber.extend(function() {
...
+        var expedite_factor = 1.5;
         this.trend = 0;
         if(this.target.altitude < this.altitude - 0.02) {
-          distance = -this.model.rate.descent * game_delta() / expedite_factor;
+          distance = -this.model.rate.descent/60 * game_delta();
           if(this.mode == "landing") distance *= 3;
           this.trend -= 1;
         } else if(this.target.altitude > this.altitude + 0.02) {
-          distance =  this.model.rate.ascent  * game_delta() / expedite_factor;
+          function getClimbRate(ac) {
```
tedrek: This function definition is ran every time the containing function is ran, wouldn't it be better to make it a method of the Aircraft object?

assets/scripts/aircraft.js
```diff
@@ -1805,14 +1805,32 @@ var Aircraft=Fiber.extend(function() {
...
+            var a = ac.altitude;
+            var r = ac.model.rate.climb;
+            var c = ac.model.ceiling;
+            if(ac.model.engines.type == "J") var serviceCeilingClimbRate = 500;
+            else var serviceCeilingClimbRate = 100;
+            if(ac.altitude < 36152) { // in troposphere
+              var cr_uncorr = r*420.7* ((1.232*Math.pow((518.6 - 0.00356*a)/518.6, 5.256)) / (518.6 - 0.00356*a));
+              var cr_current = cr_uncorr - (a/c*cr_uncorr) + (a/c*serviceCeilingClimbRate);
+            }
+            else { // in lower stratosphere
```
tedrek: For future proofing, should either throw an exception or return a usable value here.  As it is the function return is undefined above 36,152 feet.

tedrek:
The approach here looks really good to me.  If you say it's believable I'm all for it, it also comes across as a bit 'funner' to me.

assets/scripts/aircraft.js
erikquinn: @tedrek I'm still kind of trying to climb my way up the call stack and figure out where all the updates to each aircraft are actually being told to happen... I was concerned that if I made it a method, this code block would be run every time the position is updated, even in level flight, which would be an unnecessary expense to the computer. If that's not the case, I'd definitely make it a method instead.

tedrek: Methods of Aircraft are only called as needed so it's totally safe to make it a method.

module.js sets up update() to be called via [requestAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) which in turn calls aircraft_update() which calls Aircraft.update() and Aircraft.update_warning() on each frame.

erikquinn: Moved function to method of Aircraft in 65a5530.

assets/scripts/aircraft.js
erikquinn: Set the else to return the uncorrected base climb rate (obviously wrong), and marked it clearly as a placeholder in 65a5530.

Maverick283:
Looked through a couple of pilots eye videos, [here is one where you can see it quite clearly](https://www.youtube.com/watch?v=7QOCkFbRytA). 737s for example seem to try to establish 220kts until 5000ft, then on further climb accelerate to 250kts. From FSX with PMDG I know that, when the autopilot is on, the plane will climb slower to increase the speed while keeping the engines constant. Don't know if that can be noticed on a scope... 
I guess that most airplanes of a size of a 737 will behave similar, nevertheless this sound like a good thumb rule to me. I'm not sure about bigger planes, but I am guessing they are having the same speed behavior (obviously with a longer climb)

nielsvdweide:
@Maverick283 As soon as you pass thrust reduction altitude the plane will set climb thrust. It will maintain this thrust setting until you level off... As soon as you pass acceleration altitude, the plane will pitch down in order to increase speed with the same thrust setting to clean up flaps. It will then continue to accelerate to 250 knots below 10.000 feet... When passing 10.000 it will again lower the nose to increase the speed to about 300 knots and then regulate the speed with pitch as it climbs to cruise altitude. Sooner or later this will change to a mach number (about FL260) but that doesn't influence this game.

Maverick283:
@nielsvdweide That's a true Airbus fan speaking xD :D

Yeah, I agree, I thought having the plane fly 220kts when <5000ft and 250 when > 5000ft would be a easy, yet realistic procedure.

Additionally, we could introduce a value to the airplanes, that somewhat represents the climb factor. Not as in ft/minute, I'm thinking more of a value, that can be split up into climb and acceleration so it is somewhat equal to the thrust the engines can give: Let's say the value is 4000. On takeoff the engines give (ignoring flex temp/derate/ etc. for now) 100%. Thus the aircraft can either climb 4000ft per minute with constant speed, accelerate (4000/4)kts per minute with constant altitude, or climb ~3500 per minute (4000_sin(60°)) and accelerate 500kts per Minute (4000_sin(30)/4). Then, as mentioned by @nielsvdweide, reaching thrust reduction altitude (may be because of SID regulations, engine out procedures, etc) thrust goes back to 80%. Rather then 4000, the value 3200 is now used.

The angle changes with the altitude. Right after takeoff, 60° or higher should be realistic, while it is reduced to around 30° when reaching 5000. The airplane can now accelerate appropriately.

Since this model somewhat simulates realworld conditions (thrust as the one and only source for power to climb and accelerate) it should give a quite realistic climb profile. What is not taken into account here is the change in density of the air as you travel up. But that is a whole new area of physics, and I guess it should be okay to ignore when we stay under FL100.

Thoughts, ideas, comments?

@erikquinn I know this goes into a whole new direction then what your approach was, maybe just something to consider ;) Maybe it is total bull***\* what I'm suggesting here but I think it may be quite realistic...

nielsvdweide:
@Maverick283 I'm sorry, im currently in A320 type rating so yep a true airbus fan! 

The 220kt<5000 and 250>5000 is indeed an easy procedure to make this more realistic. I also think adding values to aircraft behaviour is a good thing... We could say below 5000 feet they can climb 4000ft/min and above 5000ft they climb 3200ft/min as mentioned above. This is a realistic approach to aircraft behaviour. 

I don't know why you are mentioning change of density... Aircraft climb on indicated Airspeed and that doesn't change much until switching to a Mach number. TAS does change and with the TAS change groundspeed changes as well. I don't know which speed is being used in the simulator but i think IAS will do.

Maverick283:
@nielsvdweide Density changes the turbine efficiency, doesn't it? Also, as mentioned by @erikquinn, the higher you go, the more effort you need to make to keep the aircraft flying, providing a "natural" ceiling for each aircraft... Less density -> more speed required for same updraft... And so on..

nielsvdweide:
@Maverick283 Yes and no, it depends how you look at it I guess. The only thing that changes in altitude is that the air becomes thinner and therefore you need to fly faster to keep the aircraft in the air. There is only one airspeed and that is TAS which increases with altitude to keep the aircraft in the air. But as you know every plane flies on IAS. 

In order to keep IAS the same with increasing altitude, TAS has to increase... And eventually the TAS can't increase anymore with the thrust produced and that gives the aircraft ceiling. So it more or less changes turbine efficiency, but i think it has more to do with airflow over the wing.... But anyway, we're both correct :)

erikquinn:
> In order to keep IAS the same with increasing altitude, TAS has to increase... And eventually the TAS can't increase anymore with the thrust produced and that gives the aircraft ceiling. So it more or less changes turbine efficiency, but i think it has more to do with airflow over the wing

As @nielsvdweide said is what I tried to capture: if excess power is shared between climb and acceleration at a constant ratio (eg @Maverick283's sin60-sin30), then as the aircraft approaches that max TAS, the excess thrust decreases. And the reason one must fly at a higher and higher TAS as you climb is because the air density is decreasing. So I just cut out all the middle stuff and said climb rate at a given altitude would be shaped roughly like air density, since air density causes a higher required TAS, which gets you closer to the max TAS, which is where the service ceiling is, with its +500fpm climb.

I also wanted to be sure to shape it not just for low altitude behavior, but also through the mid and high flight levels, because I see this sim including a center mode one day. I believe the current way would achieve that better than a flat model. 

And @Maverick283 I'm all for changing the plan if its a real improvement, but as I see it the two are pretty close to the same, with the one exception of the various altitudes where you pitch down and climb at a different airspeed. I'd love to see that when working the departures, but it probably doesn't add all that much in the end. For example, we could add a `Vy` property for each aircraft model, and they can climb at that until 5k, then just change their speed to 250 until 10k, then accelerate slowly to cruise-- this would be easier, require less of a rework, and probably be about the same in the end.

My thoughts anyway. If you guys really think it's worth going back and changing, that's certainly an option. There's just a long list of other stuff I'd like to get to. :smile:

Maverick283:
@erikquinn I agree that you basically cut out the middle :D Didn't think that aaaall the way through I guess... 

Since programming is always "abstracting reality", I like to program as abstract as possible. For example, "When aircraft heading is +-30° or runway heading, it may intercept the localizer" is not very abstract... It works, no doubt, yet more abstract would be "when aircraft heading intercepts radial of localizer, it may intercept" would be more abstract and simulating the whole localizer behaviour (as in simulate a signal that is sent, simulate the aircraft recieving the signal etc) would be most abstract (and probably to much) but would also be closest to reality. What I'm trying to say here is that your way will definitely deliver good results, but is only part of projection of reality. 

Maybe, once I get a bit more fluent in JS, I'll see if I can add that myself. I still think it could be a good addition in the future, especially if the airspace is extended above FL100. I was thinking it the other way around though, so in case we ever go above FL100 I would have brought a chart similar (if not the same) to yours in ;). And I would've also suggested something similar for descending: right now planes can descent and reduce the speed rapidly at the same time... I dare you try that in a plane, it is practically not possible...maybe only if the pilot deploys speedbreaks and starts slipping at the same time (not sure if aircraft structure can hold to that though...)

Enough of talking, your killer argument of cutting out the middle part and getting roughly the same result makes my suggestion go down in the deep sea of unnecessarity... Maybe the day will come where aircraft physics will have to be reworked completely :D

PS: @nielsvdweide yeah I guess we both mean the same. I'm still in favor of displaying GS rather then IAS on the Scope, which is (as you know) TAS+wind... But it'd probably confuse most ("Why is my aircraft not flying 230kts when I tell it to?")... Also engines DO get less efficient as the air is less dense -> less fuel (to keep ration constant) -> less mass accelerated out of the turbine -> less thrust... ;)

PR #361, opened 2016-01-26
New Airport: EGLL (London Heathrow)

Fechulo:
Everyone knows Heathrow but I'll do this anyways. Heathrow Airport is a major international airport in west London, England. Heathrow is the busiest airport in the United Kingdom and Europe as well as sixth busiest airport in the world. So here it is!

You can try it out here: http://fechulo.github.io/


tedrek:
Out of curiosity, the restricted area R160 appears to entirely encompass areas R157, R158, and most of R159.  Is this intentional?

Fechulo:
@tedrek If you look at [this](http://www.ead.eurocontrol.int/eadbasic/pamslight-F4EF283135E72B0724377818200C9AA4/7FE5QZZF3FXUS/EN/Charts/AD/AIRAC/EG_AD_2_EGLL_7-14_en_2015-09-17.pdf) chart, you can see that R160 actually has an unlimited altitude, which doesn't make sense because, first of all as you pointed out, it overlaps with the other areas, and secondly, no aircraft could land at Heathrow on runways 27L and 27R. I looked a FlightRadar24 and nobody seems to avoid R160, so I just set it to 2000ft like the other ones.

tedrek:
@Fechulo, interesting.  I dug a bit more and according to [this](http://www.shoeburyness.qinetiq.com/public-safety/Documents/Restricted%20Air%20Space.pdf) R160 applies to helicopters only.  It's in the case where a helicopter suffers an engine failure it must have sufficient altitude within that area to land outside of the area.  It seems to me like R160 should be just removed for our purposes as it doesn't apply to jets.

Fechulo:
That makes much more sense, I removed it in #377

PR #364, opened 2016-01-28
Set speechSynthesis language to english

erikquinn:
Resolves #338, by changing the `lang` property of the `SpeechSynthesisUtterance` from undefined to English. Now is a US-English lady controller voice. :smile:

Other issues phraseology-wise still remain, but are addressed in my forthcoming pull request.

Preview at [http://erikquinn.github.io/atc](erikquinn.github.io/atc).

Issue #338, opened 2016-01-12
Speech Recognition in German?


Fechulo:
For some reason, when I enable speech recognition in chrome, the voice is in german. This means that numbers are said in german (e.g., Delta eins zwei drei) and sometimes weird things happen in EDDM where the fixes that start with "DM" are pronounced "Deutsche Mark 1 2 3". None of my settings anywhere include german and I can't find a way to fix it.
Anyone have any ideas?
Thanks!

erikquinn:
I looked into this, but couldn't find where the voice was being specified, so I don't know if there's an English one that can be used instead. there's also issues with how the numbers are said and other phraseology inconsistencies that I am in the process of addressing. You can have a look around in my "phraseology" branch, and it should be apparent how I addressed it.

Fechulo:
I'm pretty sure it has nothing to do with the game itself, I'd say it's something Chrome's messing up somehow. I just wanted to know if you knew why this is happening and how to solve it.

harp71:
"Deutsche Mark 1 2 3" ... ROFL

erikquinn:
GOT IT!

[https://dvcs.w3.org/hg/speech-api/raw-file/tip/speechapi.html#utterance-attributes](https://dvcs.w3.org/hg/speech-api/raw-file/tip/speechapi.html#utterance-attributes)

```
var utterance = new SpeechSynthesisUtterance(textToSay); //make the object
utterance.lang = "en-US"; //set the language!!
prop.speech.synthesis.speak(utterance);  //say the words
```

Replacing the `prop.speech.synthesis.speak(new SpeechSynthesisUtterance(textToSay));` with the above seems to work! It's an american lady voice then, and the documentation seems to say there's a bunch of voices for every language, so with a little reading we might be able to get it talking correctly, AND maybe even associating particular voices with particular aircraft, which is gimmicky, but neat!

erikquinn:
This is resolved since #365 has been merged. Recommend it be closed.

engeplabs:
Still not closed yet...

PR #365, opened 2016-01-28
Phraseology Improvements

erikquinn:
Changes various pieces of phraseology to correct FAA phraseology (for the most part). A lot of work done for making the speech synthesis sound good without the need for the ui_log to display messages like `United six niner four, runway two eight right, cleared for takeoff`, and instead shows `UAL694, runway 28L, cleared for takeoff`, but reads it aloud correctly. You'll also notice that aircraft will call up to the different facilities we simulate based on the situation, eg. a departure will say `San Francisco Tower, UAL394, holding short of 28R`, and an arrival will say `NorCal Approach, BAW588 with you at 10000`. This reflects reality, as the names of the approach, departure, and tower facilities are not always the same, as pictured below.

image

PS - Sorry to the reviewer if the code in this PR is a bit scattered and difficult to rummage through. I tried to make it presentable but there were just a lot of tiny changes in a whole bunch of different places.

Give her a listen! :smiley: [http://erikquinn.github.io/atc](erikquinn.github.io/atc)

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -764,7 +764,10 @@ var Aircraft=Fiber.extend(function() {
         var length = 2;
         callsign = callsign.substr(callsign.length - length);
       }
-      return airline_get(this.airline).callsign + " " + groupNumbers(callsign) + heavy;
+      var cs = airline_get(this.airline).callsign;
+      if(cs == "November") cs += " " + radio_spellOut(callsign) + heavy;
+      else cs += " " + groupNumbers(callsign) + heavy;
+      return cs;
     },
     getClimbRate: function() {
       var a = this.altitude;
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -652,17 +652,10 @@ var Aircraft=Fiber.extend(function() {
 
       if(options.message) {
         if(this.category == "arrival" && aircraft_visible(this) && !this.inside_ctr) {
-          if(this.altitude != this.requested.altitude) {
-            if(this.altitude > this.requested.altitude) var alt = "descending through "
-              + radio_altitude(this.altitude) + " for " + radio_altitude(this.requested.altitude);
-            else var alt = " climbing through " + radio_altitude(this.altitude) + " for "
-              + radio_altitude(this.requested.altitude);
-          }
-        else var alt = "at " + radio_altitude(this.altitude);
-          ui_log(airport_get().radio.app+", "+this.getRadioCallsign()+" with you "+alt);
+          this.callUp();
           this.inside_ctr = true;
         } else if(this.category == "departure") {
-          ui_log(airport_get().radio.twr + ', ' + this.getRadioCallsign() + ", ready to taxi");
+          this.callUp();
         }
       }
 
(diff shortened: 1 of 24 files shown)


Maverick283:
i love it!!!

BUT: everytime i try to assign an altitude, ot says altitude not understood... maybe that is because im on my surface, which i doubt, or i am just having one of these days  where nothing works, which is more likely, or it has something to do with the changes....

also she pronounces wind as "waind", as you would say in reWIND, which is weird...

Maverick283:
is there a way to give the waypoints an additional attribute, so that she pronounces them correctly??? like ARCHI for example, she spellt that out in phonetic letters rather then saying it...

erikquinn:
@Maverick283 I took another look, and it is printing and reading altitudes back to me properly. Maybe try an incognito chrome tab (so the cache is definitely cleared of old stuff) on a real computer. If that works, and the surface still doesn't, that might be a separate bug itself.

But yes, I too noticed that she says "waind" as in "winding a clock", but it's only about 30% of the time in takeoff clearances. I couldn't figure out what was causing her to choose between the two pronunciations, but with fresh eyes I just now figured out to just change `wind` to `wynd` in the _pronounced_ sentence only, and that seems to have fixed it (see 6300b9d).

As for the fixes, she tries with some, and just spells out others. I'm not really sure there, that one will be more tough, but seems like a problem for the future.

Changes are live at [http://erikquinn.github.io/atc](erikquinn.github.io/atc)

Maverick283:
Okay, on the laptop altitude assignment is no problem! Will see if I figure out what caused it on the Surface, but this has obviously nothing to to with this pull request ;)

Wind is pronounced fine now, good idea to change it!

Something else I struggle understanding right now: Even though this is not merged yet, when I play at http://zlsa.github.io/atc/ I can enable Speech output already, BUT: It is in German! This is actually quite funny as she has that unbelievable strong german accent, yet it is not how it should be. 
Any ideas what could have caused this?

Maverick283:
As mentioned in #368, there seems to be a issue with the "heading" a.k.a. "h" command.

It must be related to @erikquinn's branch as it only happens when playing on his(your) version.

[See a video here](https://youtu.be/C-gCQQxJuek)

erikquinn:
@Maverick283 it has been a feature for a pretty long time, and has always had the German accent, which led to weird pronunciations like mentioned in #338, which are fixed in PR #364.

**Hold this PR** while I look into the heading issue.

erikquinn:
Resolved the turn issue in 48f3aeb, and edited some old commits for tiny fixes (eg apparently KSAN tower is called "Lindbergh Tower", which I didn't bother to look up, a spelling mistake, and similar unimportant changes).

Due to there being a conflict between this and #364, I merged #364 into this PR, applied the intended merge resolution, and then closed #364.

Gave it a more thorough test this time (which led to all the nitpicky changes). Should now be good to go. Test if you like at [http://erikquinn.github.io/atc](http://erikquinn.github.io/atc) (fyi my gh-pages also has #356 merged in, relating to more realistic climb performance).

#### Ready to merge

Maverick283:
I agree, this is looking and working good!!! Didn't find any more bugs, good work @erikquinn!

assets/scripts/aircraft.js
```diff
@@ -996,7 +1003,12 @@ var Aircraft=Fiber.extend(function() {
 
       if(response.length >= 1) {
         if(response_end) response_end = ", " + response_end;
-        ui_log(this.getRadioCallsign() + ", " + response.join(", ") + response_end);
+        
+        var r_log = function f(){var x = [];for(var i=0;i<response.length;i++){x.push(response[i].log)}return x;};
+        var r_say = function f(){var x = [];for(var i=0;i<response.length;i++){x.push(response[i].say)}return x;};
```
tedrek, line 1008: This should be writeable as:

```
var r_log = response.map(function (e) {return e.log}).join(", ");
var r_say = response.map(function (e) {return e.say}).join(", ");
```

harp71:
Nice, one little bug though: if you enter a wrong command there´s a double comma with a space between.
image

erikquinn:
@harp71 Good catch. Will be fixed in #383. :smiley:

PR #377, opened 2016-02-06
Remove restricted area "R160" at EGLL

Fechulo:
Thanks @tedrek for pointing out that R160 applies to helicpoters only!


tedrek:
:+1:

Issue #384, opened 2016-02-08
Circular or Sector-Shaped Restricted Airspace


siyuanr:
This is a partial extension of the discussion in #219. Some restricted areas are defined using a center location, radius, and (if it is a circular sector), the range of degrees. Can we implement this?

In the following example, in the vicinity of `VHHH`, `VHD5` is a firing range defined by coordinates; `VHR12` is an entertainment park with a circular shape, radius 1.35NM; `VHP8` is a satellite station whose restricted airspace consists of two circular sectors with the same center.

<img width="594" alt="screen shot 2016-02-08 at 06 08 58" src="https://cloud.githubusercontent.com/assets/15655087/12878796/a4fb3554-ce2a-11e5-9c8a-51e21ca76aab.png">
(Source: ENR 5.1, AIP Hong Kong)

sixeyeco:
It's already implemented, a few airports have airspace restrictions. On my fork, at an airport called M002 I have a single restricted circle to simulate a police closure with a 3k foot ceiling. 

I've been wanting to post a wiki page showing how to create these zones quickly using a website and an excel spreadsheet. With accurate GPS coordinates it's a snap. If "close" is "good enough" then you can get general shapes with a point and click solution. 

I found the coordinates at http://www.ais.gov.hk/HK_AIP/AIP/ENR/HK_ENR5.1.pdf. If you confirm that they are correct then I will write the code for you and detail the steps I took.

sixeyeco:
I generated the restriction for VHR12 at VHHH. You can test it on [my fork](http://sixeyeco.github.io/atc/). 

[Here is the link to the raw airport file if you want to copy it.](https://raw.githubusercontent.com/sixeyeco/atc/gh-pages/assets/airports/vhhh.json)

siyuanr:
@sixeyeco Excellent! Could you show me how to extrapolate the circumference coordinates based on center coordinate and radius (presumably in Excel)?

sixeyeco:
Sure. I use [this site](http://www.gpsvisualizer.com/calculators). Enter the radius and center coordinate in the section titled "Draw range rings around a point" and click "show coordinates". That might be all you need if you have more experience at coding than I do.  

[...]

Issue #370, opened 2016-01-29
New SIDs, Fixes, and other minor problems for VHHH


siyuanr:
According to the charts [published by the HKAIS](http://www.ais.gov.hk/HK_AIP/ad.htm) (a government agency), we are missing some SIDs and important fixes.

Fixes that needs to be added for new SIDs:

``` json
"ATENA": ["N22d24m39.85", "E114d23m11.88"],
"COLEY": ["N22d06m41.03", "E114d18m37.63"],
"RAMEN": ["N22d09m39.22", "E114d05m09.89"],
"PORPA": ["N22d20m09.10", "E114d01m16.30"],
"ROVER": ["N22d20m35.58", "E114d01m39.12"]
```

Could someone add those to `vhhh.json` even if what's described below couldn't be done?

Now comes the tricky part, which is in some way an extension of Issue #324 in `VHHH`. However, in Hong Kong's case, paths with the same end fix but different departing runways are treated as different SIDs, which are given different names, and there may be more than one SID between a runway and an end fix.

For example, for the same `BEKOL` SID, the HKAIS has, for RWYs 7L, 7R, 25L, 25R, respectively:

``` json
"BEKOL 3C": ["ROVER", "TD", "ATENA", "BEKOL", "IDUMA"],
"BEKOL 3A": ["PORPA", "TD", "ATENA", "BEKOL", "IDUMA"],
"BEKOL 2B": ["PRAWN", "RUMSY", "TUNNA", "SAMON", "BEKOL", "IDUMA"],
"BEKOL 2D": ["PRAWN", "RUMSY", "TUNNA", "SAMON", "BEKOL", "IDUMA"]
```

Will this cause any problems? If not, we could then proceed to write the different versions of the other SIDs.

We may also consider adding `ATHENA` SIDs for RWYs 7L, 7R, since the HKIAS has:

``` json
"ATENA 2C": ["ROVER", "RAMEN", "TD", "ATENA"],
"ATENA 2A": ["PORPA", "RAMEN", "TD", "ATENA"]
```

However, since `ATHENA` is still far away from the radar boundary, we may need to come up with an extender fix further to the east, just as what `IDUMA` is for `BEKOL`. Also the number of destinations may need to be expanded.

[...]

PR #385, opened 2016-02-09
Added all the restricted areas to VHHH

siyuanr:
Please refer to discussions #384 and #370.


siyuanr:
Brief comment on Guangzhou FIR ZGGG AR05. The airspace to the north of VHHH is controlled by Guangzhou FIR. Mainland China requires that incoming flight from Hong Kong FIR muss be handed off to Guangzhou FIR at or above 4,800m (FL157, as [confirmed by the HK Government](http://www.info.gov.hk/gia/general/201410/15/P201410150241.htm)). In my opinion Guangzhou FIR can therefore be treated as a restricted area.

tedrek:
@siyuanr, thanks for the updates.  Implementing Guangzhou FIR as a restricted area makes sense to me, however it looks like you have reverted.  Either update this PR or create a new one if you would like to re-implement it.

siyuanr:
@tedrek The problem is that (even a sector of) Guangzhou FIR is disproportionately big for the map.  As a result the FL restriction, written in the center of the FIR, is placed far away from map center. Another thing is that the restriction only applies to the border crossing -- cross it, and you follow whatever the height the Guangzhou FIR controller directs you to. The FIR border should be better implemented as a line in a future geojson file, or in an added feature of `airport.js`.

As a stopgap measure I could delete the waypoints and navaids belonging to Guangzhou FIR, so that the airspace of Guangzhou FIR would appear to be mostly empty. Do think it's good?

PR #386, opened 2016-02-11
Complex SIDs

### LINKED ISSUE #324: SID/STAR not sensitive to runways
image
The above behavior for the OFFSH9 is undesirable, and stems from the sid being defined simply as a list of fixes, "SEPDY, SENZY, WAMMY, SEGUL", when the reality is infinitely more complex. Though not all of that complexity is necessary to capture for our purposes, it seems now to be a good time to take a step closer.

---

offsh9
This is the chart for this particular sid out of SFO, and you can see based on the arrows that come out of the airport that the behavior is different based on the runway an aircraft departs from, and logically so.
- a/c departing _NORTH_ fly hdg 333 to intercept the SFO R-350 to SEPDY, then turn left hdg 203 to intercept the PYE R-151 to SEGUL
- a/c departing _WEST_ fly runway hdg to SENZY, then turn left hdg 203 to intercept the PYE R-151 to SEGUL
- a/c departing _SOUTH / EAST_ cannot use this SID (they would be assigned a new one in the event of a wind shift that caused a runway change)

---

north dep
Now, having the a/c follow the sid AS PUBLISHED would result in something that looked more like this^ (he would also turn in line with the dashes that extend SE of WAMMY). Sadly, this accurate and "correct" behavior is not in confirmity with the clear implications made by the dashed lines that show the SID path, and could surely confuse some users. Whether it's worth going to the realism of intercepting radials/fixes I'm not sure; maybe it's a more attractive compromise to just make them go to the right fixes for each runway. Looking for feedback on that topic.

I'd like to start work on this issue today... feedback anybody? :smiley:

erikquinn:
Adds the ability to have more flexible and capable SIDs. Arrays of fixes can be included separately for each individual runway of a SID, the main body of the sid, and then any SID transitions. Thus, complicated SIDs are now possible to implement under the same name, eg `OFFSH9`, and the aircraft `fms` will fly to the appropriate fixes for whichever runway and transition they may have chosen, and all we have to do is assign them the SID. It would be easy to expand this to implement other things, such as speed restrictions, inbound/outbound radials (once the `fms` can handle them), and other such complicated stuff if desired later on, and will translate well to STARs, though those are currently not as important to us operationally.

---

Most of the lines changed (1262 of 1557) are simply from writing out the SIDs in the airport files, so it's not as huge a commit as it may look. I spent some time looking over the charts, but did not add altitudes as I saw on the charts yet. Maybe some time I will come back and get all the SIDs for all the airports fully outfitted, but that could take a while anyway. At some point, I will get the USA airports up to date at least. For now, I simply updated all airport sids to the new format, and kept all the old stuff intact (except some minor renaming).

It's also worth pointing out that adding all the complexity and additional fixes and whatnot made it so that the blue sid lines basically end up connecting absolutely everything on the screen.... I cleaned them up, but even when drawn properly, the complexity made them look atrocious, and so I chose to remove them. The path of a particular sid is still shown when an aircraft is assigned it of course, but the little airplane menu button is gone, as are the blue text labels at the end of each SID. Also, the blue arc on the edge that defined where the a/c was supposed to exit was removed, as it is no longer relevant. At some point, it should be replaced to verify that the a/c was cleared onto its route eg via `f`, `pr`, or `sid` commands. Just FYI, as it is something to consider when looking this over.

---

image
Also includes:
- change of `airport.destination` from heading basis to based on the SID the a/c filed for
- addition of `caf` command ("cleared as filed"), to clear an aircraft to an `initial_altitude` (different for each airport. defaults to 5000), and to the destination via the filed SID and route.
  - addition of `cvs` command ("climb via sid"). The fms will be filled with the appropriate altitude restrictions as charted, and the aircraft will climb to the ceiling of the airspace, while complying with the altitude restrictions.

**Resolves #324**

Test: [erikquinn.github.io/atc](http://erikquinn.github.io/atc)


Maverick283:
I like it a lot! Except for that you removed the arcs... (especially when I'm new to the airport) I have no clue what direction the airplane will be heading... This is important for when I have two aircrafts takeoff on parallel runways shortly behind each other, as it I'd like them not to cross their paths but to fly away from each other... 
As a real world controller you know the sids by heart, we don't though... The approximate point the aircraft will leave the sector at is only a small information (I honestly would like to see the whole path before I clear them for takeoff), I guess it is enough to work with though

erikquinn:
Yeah I understand. And just because I study ATC doesn't mean I know the SIDs for all the airports; I surely don't. But since I can see the full path as soon as they have a takeoff clearance, I can live without the blue arc for now. By changing the `destinations` to be based on a SID instead of a heading, it makes it tougher to draw an arc. But after playing for 10 minutes on a new airport, I found that I learned what SIDs went where, and as a player, it made me feel proud to be learning the airspace on my own. 

That said, I agree that we need at least an option of making it more apparent where each SID goes. I just couldn't get an easy solution, and figured it wasn't critical for now, and could be added later. 

This PR is definitely somewhat "rough around the edges" in terms of stuff like this, so it will take some more work to decide on solutions to minor issues like this, but I didn't want to hold back the pull any longer as I believe at this point the benefits outweigh the few issues like the blue lines/arcs.

siyuanr:
How should we change the map in order to contribute to your work? I mean, do we file a separate pull request, or do something else?

siyuanr:
Also, what happens when an airport doesn't have SIDs (such as `KORD`, which I'm working on)? Is the arc still shown? At `KORD` all planes expect direct to first waypoint on NAVAID without flying an SID.

Maverick283:
@erikquinn I agree that I too learned it quite easily. Maybe that is another one on the bullet list of options: "Make SID-paths visible". Rather then having them on the map all the time (which I liked too, yet I can understand it can get messy!) or having them displayed after takeoff, what if as soon as the sid is assigned all variants of it (depending on the runway) are being displayed? This should make the decision which runway an airplane has to be sent to to make it the most efficient easier?

erikquinn:
@siyuanr I'm not sure how to go about that.. maybe merge my changes in, and just link a comment in this issue to your branch? And if we like the combination, make a PR with your changes, and note that it depends on this being merged? That could probably work...

@Maverick283 yeah if you remember there is actually that little menu button, so it could stay there, defaulted to off, instead of going in the settings menu. And "having them displayed after takeoff" is what it currently does; if you clear an a/c for takeoff and highlight it, the sid path is drawn until the a/c is deselected. So we aren't totally out of ways to see where a sid goes.

@all I pushed a separate branch to my fork that includes this pr AND the blue lines. You'll notice it looks okay at most airports, but the ones that have the SIDs more thoroughly and accurately implemented (like KSFO) appear much more cluttered than I would like my scope to look. Here's an example of why I contend that it does more harm than good:

image

---

You can view this PR _with_ the blue lines here:
[http://erikquinn.github.io/atc/b/complex_SIDs_bluelines/](http://erikquinn.github.io/atc/b/complex_SIDs_bluelines/)

Maverick283:
@erikquinn Yeah, after I have assigned them a SID (which I don't exactly know where it will take them) AND a runway (which makes it even harder for me to know where exactly they are going to go, even if I know where the SID terminates) AND cleared them for takeoff (so I cannot "undo" the commence of their flight path) only THEN I can get the flightplan displayed. What I was suggesting is that as soon as they have their SID assigned (when they are still on the ground) that all possible SIDs for the runways (maybe even only the active ones) are being displayed... So that one can pick the runway based on the SID path.

assets/scripts/aircraft.js
```diff
@@ -714,16 +753,6 @@ var Aircraft=Fiber.extend(function() {
...
-        // Fly away!
-        this.fms.setCurrent({
-          navmode: "heading",
-          heading: this.radial,
-          turn: null,
-          hold: false,
-          altitude: 20000,
-          speed: 330,
-        });
-
```
tedrek, line 726: The intent of this code was to ensure an aircraft outside of control would always leave the game.  For example an aircraft can be in fix track mode but end up circling the fix indefinitely without a guard like this.

assets/scripts/aircraft.js
```diff
@@ -1162,6 +1199,54 @@ var Aircraft=Fiber.extend(function() {
...
+          }
+          else if(a.indexOf("-") != -1) {
+            var maxAlt = parseInt(a.replace("-",""));
+            var alt = Math.min(alt, cruise) // climb as high as restrictions permit
+          }
+        }
+        else var alt = Math.min(cruise, airport_get().ctr_ceiling);
+
+        wp[i].altitude = alt; // add altitudes to wp
+      }
```
tedrek, line 1240: Stylistically, this sort of manipulation of FMS structures means the code should probably be part of FMS instead of Aircraft.

tedrek:
@erikquinn, some really awesome work here.  The approach looks flexible

One request, could you write some documentation for it?

Regarding drawing SIDS, have you tried drawing just the bodies/transitions without the per runway section?  I definitely think having some sort of visible portion is necessary, but I totally agree showing all of them is too much.

The one other thought I just had (and don't have time to look at) is how these departures are scored if they leave the 'wrong' directon.

assets/scripts/aircraft.js
erikquinn, line 726: Understood. I was removing it because it would then peel the departures off of their flight plan when flying a SID, and just put them on a heading away from the airport. It looks better to just leave them alone, and allow them to continue along the SID path. Of course, it would be possible that aircraft may not always leave the game, but after adding SIDs, the only time this block of code would be needed (that I can think of now) would be if they were told to fly to a fix outside of the airspace, and then another inside the airspace.

So I actually think when SIDs are implemented, this should be removed.

assets/scripts/aircraft.js
erikquinn, line 1240: True. I'll fix that.

assets/scripts/aircraft.js
```diff
@@ -334,10 +334,16 @@ var Model=Fiber.extend(function() {
 zlsa.atc.AircraftFlightManagementSystem = Fiber.extend(function() {
   return {
     init: function(options) {
-      this.aircraft = options.aircraft;
```
erikquinn, line 337: @tedrek This is an interesting one.

I removed the aircraft object from being a property of the fms object due to the infinite inheritance (eg. aircraft has an an fms, which has an aircraft, which has an fms, etc). Is there some way to stop this? Or maybe we need a `get_aircraft()` function like the `get_airport()`? When I move the SID-flying logic from aircraft to fms, I will need to be able to reference the aircraft, and it seems like `fms.aircraft` would not be the way to go because of the issues above. Do you follow/agree/disagree/have feedback?

erikquinn:
@tedrek Thanks. Unfortunately, as I came to learn more about the european style SIDs (apparently they all go to a single fix, it made the format I used in the SIDs excessively long and complex, and lacked the ability to call the sid by its proper name, eg `EKERN 2C` instead of just `EKERN`... among other things, I actually think I want to close this down and rework a few things. Better to get it right than half right, I always say.

I'll be happy to write up some documentation. Since we aren't using the wiki for developers, is there a dev-wiki we want to add where the operation of this sort of thing can be described?

And yeah, I've tried drawing just those, but the variation between each DP doesn't lend itself toward that working very well. For example, the european single-fix SIDs have most of their content in the runway sections, and merge like 20+ miles away from the airport, which isn't super helpful. But maybe; i'll definitely give it another look.

As for scoring, I just straight-up removed it for now (not a good solution). The ideal answer would be to get points based on whether or not they were cleared onto their flight plan route before they left the airspace. With some airports still not having SIDs fully and properly implemented (still using headings like `110` as destinations instead of fixes at the end of a SID, I can't make the switch to grading based on clearing them onto their route yet. I see the workflow like this:
- Complete SIDs logic (with a PR like this one, but less bad)
- Add/fix all SIDs for all airports, complete with speeds/altitudes
- Make scoring based on whether or not a/c left on their flight plan route

---

**Anyone who was waiting for #324, go ahead and continue using the current system. It will be another few days before I can get this the way I want it.**

Closing this now for additional work.

assets/scripts/aircraft.js
erikquinn, line 1240: Fixed in new branch (complex-SIDs2).

assets/scripts/aircraft.js
erikquinn, line 337: Fixed in new branch (complex-SIDs2).

I left it as removed, and created a new property for each aircraft, an "entity id", or "eid". This is set at spawn (and shifted as planes are removed) to be the index of a given aircraft within `prop.aircraft.list`. This makes it VERY convenient and easy to access different aircraft regardless of scope. When the fms is initialized, it also keeps a property `my_aircrafts_eid`, which can be used to refer back to the fms's parent object: the aircraft itself.

PR #396, opened 2016-02-18
Localizer interception

### LINKED ISSUE #375: Localizer Intercept Angle
As I alluded to briefly in #352, I was "pretty sure" that we weren't handling the approach course interception rules correctly in the code. The applicable paragraph out of the [65](https://www.faa.gov/documentLibrary/media/Order/ATC.pdf) is para 5-9-2, and provides a table indicating the normal intercept angles, 30 degrees most commonly being the applicable offset.

So, in the code, a la line [1637-1647](https://github.com/zlsa/atc/blob/b4b266c4177d7f8ee5f727e70e20b0125d38e3c9/assets/scripts/aircraft.js#L1637-L1647):

```
//plane is on the glide slope
if(this.mode != "landing") {
    this.mode = "landing";
    if (!this.projected &&
        (abs(angle_offset(this.fms.currentWaypoint().heading, angle)) > radians(30)))
        {
            ui_log(true, this.getRadioCallsign() +
                       " landing intercept vector was greater than 30 degrees");
            prop.game.score.violation += 1;
        }
```

We calculate the difference between assigned heading and runway heading, and see if it's over 30. The catch (which I have verified today with a retired controller and trainer at the FAA Academy) that yes indeed, those 30 degrees are based not on the runway's precise magnetic heading, but on its nominal heading.

For example, SFO's 28R & 28L have a magnetic heading of 284, but are called "28" left/right, and it's the "28" part that those headings are based off of. So, even though 250 is 34 degrees offset from runway heading of 284, it is allowable because it is 280 - 30 degrees.

The logic should be changed to compare assigned heading not with the literal `angle` of the runway, but the _nominal_ angle, eg. 280 degrees.

erikquinn:
Resolves #375 and also fixes another issue I never wrote up an issue for... when a/c are given a heading, they are expected to fly that heading until joining the localizer. In the sim, they kind of just fly it until they get close-ish to the localizer and then turn TOWARD the approach course, which would not be allowable, and makes it difficult to control the sequence because you are forced to withhold approach clearances to prevent the undesirable behavior. Now they will actually stay on the heading until they need to turn onto the localizer.

See pictures below. Both aircraft were assigned a heading of 275, which they have been flying for a few miles. You'll notice how EL turns left to a sharper intercept angle, while the speedbird stays on the assigned heading all the way to the localizer.

before:
image

after: 
image

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -1673,16 +1673,11 @@ var Aircraft=Fiber.extend(function() {
       if(this.fms.currentWaypoint().navmode == "rwy") {
         airport = airport_get();
         runway  = airport.getRunway(this.fms.currentWaypoint().runway);
-
         offset = runway.getOffset(this.position, this.fms.currentWaypoint().runway, true);
-
         offset_angle = vradial(offset);
-        
         this.offset_angle = offset_angle;
-
         this.approachOffset = abs(offset[0]);
         this.approachDistance = offset[1];
-
         angle = runway.getAngle(this.fms.currentWaypoint().runway);
         if (angle > (2*Math.PI)) angle -= 2*Math.PI;
 
@@ -1701,7 +1696,6 @@ var Aircraft=Fiber.extend(function() {
         if ((abs(this.altitude - glideslope_altitude) < glideslope_window)
             && (abs(offset_angle) < radians(10))
             && (offset[1] < ils)) {
-          //plane is on the glide slope
           if(this.mode != "landing") {
             this.mode = "landing";
             if (!this.projected &&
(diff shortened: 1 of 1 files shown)


MaicomMR:
Wow, really very very cool man! :+1:

harp71:
@erikquinn makes sense, nice!

sixeyeco:
Much better. Well done. :+1:

erikquinn:
Possible bug with this. Hold the pull for now please.

erikquinn:
Fixed the bug... it was small and easy. Sorry it took so long to get to it; I was in the process of moving about 1,000 miles from Akron, Ohio to Moore, Oklahoma. Finally all set up here and came back to fix this. This is now ready to merge.

Issue #353, opened 2016-01-21
"waiting" on the world to change


erikquinn:
![image](https://cloud.githubusercontent.com/assets/5103735/12492539/aeade28e-c04e-11e5-987e-8b9106aa87c1.png)

Not to draw parallels to John Mayer with my title :stuck_out_tongue:... but _what on earth_ are all these departures waiting for? Most in the screenshot were taxiied as much as ten minutes ago, and they just never seem to pop up at the runway ready to go.

Getting that taxi time reduced would also help with the strip organization issues brought up in #340.

sixeyeco:
I run into this at least once per session. I restate the entire command to taxi (alt, speed, runway) and they appear. They won't jump the line (they queue up as if they only just received clearance) but it gets the job done. 

I've always assumed that I accidentally assigned it a fix or SID without ever directing it to taxi. But now that I see you're experiencing the same thing I'll look back at some old footage to verify that.

tedrek:
I have never seen aircraft order to taxi which don't end up at the runway.

However, there is an odd timing behavior which you may be running into.  If an aircraft is ordered to taxi to a runway and no aircraft is waiting then it will take 'delay' seconds plus zero to three seconds to arrive on the runway.  At SFO this means an aircraft takes from 2-5 seconds to show up after you issue a wait command.  If there is already an aircraft waiting to show up, then the delay becomes 60 seconds from the time the next aircraft appears.  Could this be the behavior which you are noticing?

This does seem backward to me, having a line of aircraft waiting for the runway should mean they are ready for takeoff much quicker than having to taxi from the apron.

erikquinn:
The issue is probably related to that, it sounds kinda overly-complicated... But not quite even, its much more than 60s. It only happens occasionally, but when they skmrwhow get stuck like this, nobody shows up at the runway ever. 

If it helps, using other runways will still work, until that one gets jammed up too. And they all seem to be getting there, because they will tell you who they're waiting behind.  Maybe it's just a problem with the first guy in line not showing up, but still being ready to go. It's just tough to figure out who is first. I suspect that if I cleared the first in line of the pileup for takeoff, the subsequent ones would continue appearing as normal, but I have yet to verify this; its just my hunch.

harp71:
@erikquinn this might be related to what i found out here #105

erikquinn:
@harp71 oooh definitely-- good catch. that will be very helpful to track down what's causing it.

erikquinn:
Resolved by #400.

PR #400, opened 2016-02-20
Clear Runway After Crash

### LINKED ISSUE #105: Error makes it impossible for planes to takeoff
screen shot 2015-05-26 at 3 35 26 pm

Two planes on the runway (though the UI doesn't show them anymore). 
Commands to takeoff are ignored. Looks like a JS error. Perhaps #19 could fix this.

erikquinn:
Resolves #105 and #353.

When an aircraft who is still in the takeoff queue for a given `runway`, and has a collision with another aircraft, both are removed from the _game_, but the departure is not removed from the runway's queue. This fixes that. Now if a collision takes place, little cyber ninja monkeys scurry out onto the tarmac and sweep up all the crash debris so the next departure can get on with its life (they're fast too-- the runway is available within milliseconds!) 

Highly technical explanation in emoji speak:
:boom: :airplane: :boom: :fire: :fire_engine: :heavy_plus_sign: :monkey: :arrow_right: :airplane: :+1: 

Shout out to @harp71 who figured out how to reproduce it. Made it pretty easy to fix, so thank you!

---

Demo: [http://erikquinn.github.io/atc/b/sweep-the-crash-debris/](http://erikquinn.github.io/atc/b/sweep-the-crash-debris/)

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -108,6 +108,18 @@ zlsa.atc.Conflict = Fiber.extend(function() {
         prop.game.score.hit += 1;
         this.aircraft[0].hit = true;
         this.aircraft[1].hit = true;
+
+        // If either are in runway queue, remove them from it
+        for(var i in airport_get().runways) {
+          var rwy = airport_get().runways[i];
+          var queue = rwy.waiting[0];
+          for(var j in queue) {
+            if(queue[j] == this.aircraft[0])
+              rwy.removeQueue(this.aircraft[0], rwy, true);
+            if(queue[j] == this.aircraft[1])
+              rwy.removeQueue(this.aircraft[1], rwy, true);
+          }
+        }
       }
     },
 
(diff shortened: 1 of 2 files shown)


harp71:
THX @erikquinn 

Plz tell me one thing, because i been told it´s not possible to review a branch other then gh-pages on the github website: how did you get this URL for the demo?

http://erikquinn.github.io/atc/b/sweep-the-crash-debris/

I mean, it´s logical and almost self explanatory exept .../b/... :)

erikquinn:
@harp71 it's kinda painful to do, but i would copy all of the files from the desired branch to a folder. Then I'd checkout gh-pages, and add a new folder "b" (as in branch), and then added a subfolder within b that was the branch name and copied all the files in there. Then I'd commit "added 'branchname' branch to subfolder" and push, and repeat the process for as many branches as I wanted to be testable concurrently. 

The only huge headache was that sometimes git didn't recognize me adding the files in `/b/`, and eventually found on stackexchange that it was being ignored, and I had to add `~/*` or some similar nonsense to the `.gitignore` so it would recognize me making the changes to the untracked folder "b". 

Overall, it's kinda annoying, and has to be completely redone each time ulyou make an update to a branch, but definitely a nice option to have.

Issue #399, opened 2016-02-19
SBGR Runway & Fix Locations


erikquinn:
I'm not sure why, but it looks like the runways and fixes at SBGR are not in their correct positions. I noticed this when I went to add two new fixes (IG and PIER)... they were not where they I expected them to be, relative to the airport. Turns out the airport is actually placed 15 miles away from where it should be, and the fixes are shifted to match.

In the below picture, you can see this. I moved runway 9L-27R to the correct position, and also moved CGO to the correct position (old one is still there, labeled "CGO-OLD" for demonstration purposes).

![image](https://cloud.githubusercontent.com/assets/5103735/13188432/90419196-d71f-11e5-8658-e60c795ca383.png)

To make future elaborations on this airport easier, this should be corrected to use the real-world positions.

MaicomMR:
Your runway positions is in the wrong position.

This is the airport view:
![untitled-1](http://photos.airliners.net/aviation-photos/photos/6/3/0/0988036.jpg)

I'm live on Brazil and fly to São Paulo some times, I make this summary of city airports.
![untitled-1](https://cloud.githubusercontent.com/assets/12571488/13191365/5032ad6c-d74a-11e5-957b-98f11d9a94d1.png)
SBGR > GRU Airport is the international Airport
SBMT > Campo de Marte - small planes, private and police helicopters base
SBSP > National Airport - Flights to all brazil, its the Urban Carrier, only small planes can land on them(A318, A319, B738)
The CGO waypoint is above the SBSP airport.

When I rebuild the SBGR I use official AIS(Aeronautical Information System,) charts, with real runaways, waypoints and SID coordinates, but someone change the airport after.

You can found official charts here:
http://www.aisweb.aer.mil.br/?i=cartas&filtro=1&nova=1

Sorry my bad english.

erikquinn:
@MaicomMR You are right... they _were_ correct, until commit cd6c496, when they were changed to the positions in the bottom picture (middle of the forest). I'm not sure why this was done, but since all of the other fixes were moved as well, I'm assuming that was meant to move them to the new airport position. Strange.

What it used to be, when @MaicomMR first added it:
![image](https://cloud.githubusercontent.com/assets/5103735/13192456/2277a2f2-d73a-11e5-8a27-cd144b281fd2.png)

What it changed to:
![image](https://cloud.githubusercontent.com/assets/5103735/13192461/32ce72a2-d73a-11e5-84e4-5197d140cfe8.png)

[...]

PR #404, opened 2016-02-22
Complex SIDs

### LINKED ISSUE #324: SID/STAR not sensitive to runways

erikquinn:
Resolves #324, #399.

This monstrosity is the follow-up to the now-closed PR #386. (See there for thorough description)

Improvements since #386:
- SID blue lines fully restored. They do look complicated at airports with mature SIDs however!
- "climb-via-sid" and "cleared-as-filed" moved to the fms
- Simplified SID format in airport json, made more flexible so it makes sense both in european-style (fix-based) and FAA-style (fix-transition-based) SIDs.
- Updated ALL airports' SIDs (some to the real charts, some just to the new format)
- Airports still using headings for departures had to be changed to placeholder SIDs (north/south/east/west, for example see KMSP)
- Updated departure scoring system (now based on whether or not the aircraft has passed or been cleared to the last fix of the SID in their flightplan (`fms.fp.route`).
- Updated documentation on new SID format and a few other relevant changes to documentation
- Updated tutorial to teach to use SIDs
- cleaned the crap out of the code, and rebased/amended until it actually became organized and clear

Available live at [http://erikquinn.github.io/atc/](http://erikquinn.github.io/atc/). :smile:


MaicomMR:
Wonderful, you are a genious my friend :+1:  :+1:  :+1:

PR #413, opened 2016-03-08
Update EDDH with new SID's

Fechulo:
-Updated SID's with new format
-Added SID's to all runways
-Updates to some fix names
-added magnetic heading

PR #415, opened 2016-03-09
Update EGLC with new SID format

Fechulo:
-Update to new SID format with correct SID's for all runways.
-Added new BPK SID
-Added magnetic heading
-Added new STAR's (EGLC has recently been sold and a LOT of changes have been made to the STAR's)
-Added restricted areas


assets/airports/eglc.json
```diff
@@ -76,52 +942,67 @@
     "DVR": {
       "icao": "DVR",
       "name": "Dover Five",
-      "suffix": {"09":"", "27":""},
+      "suffix": {"09":"5U", "27":"5T"},
       "rwy": {
-        "09": ["GINTI", "DET", "DVR"],
-        "27": ["GINTI", "DET", "DVR"]
+        "09": [["LCE01", "A570+|S210"], ["LCE02", "A3000|S210"], "BEMID"],
```
erikquinn: Follow-up to inline comments on #413: In the unusual situation like this where there actually exists a restriction of 570 feet (usually I haven't seen precision exceed 100ft increments except very rarely), you can't express it with full precision in flight levels (either FL005 or FL006). But the way the logic is implemented, you can preserve the precision by simply entering `"A5.7+|S210"` instead of `"A570+|S210"`. It will multiply out properly.

erikquinn:
In testing, I was getting some freezing on this branch every once in a while when using the `caf` (cleared as filed) command, with this airport only. I'm not sure what the deal is, but there is a problem within the `fms` trying to invoke `fms.addWaypoint()`, and some reference error because something is passing invalid data. I can't seem to make sense of it, nor do I have any theories on why it seems to be specific to this airport.

@Fechulo Are you observing any freezing on this branch?

Fechulo:
I don't recall experiencing any freezing, but I'll test it now and I'll see if I can find the cause.

Fechulo:
I found the bug! After an extensive 5 minutes of testing I realized that the game only froze when the "caf" command was given to aircraft requesting the Brookmans Park (BPK) departure. After a quick look at the code, I found an sneaky "a" hiding behind the waypoint "LCN01". It is now fixed.

erikquinn:
Ah, that would do it. haha thanks. I'll get to this tomorrow.

PR #424, opened 2016-04-06
Improve Separation Logic!

erikquinn:
This PR includes somewhat hefty back-end changes to the ways Runway objects, their methods, and their relationship as a child of Airport works.

Basically, each piece of pavement is treated as one "runway" and to get Runway 18 as opposed to Runway 36, you have to use the `.getEnd("18")` method. This is only an inconvenience, it was really some issues with data being associated with the "runway", and not the "end" (for example, the runway's `angle` is the heading from the primary side, and there is NO way to ask for the heading of a runway that you're working with... you would have to find out if you have the right side, and if not, add `Math.PI`, because it's facing the wrong way.

It would make a lot more sense to have a `Runway`, which has two `end`s, each of which have their own `angle`, `length`, `name`, etc, so you could get the runway's name as simply as `airport_get().getRunway(rwy).angle` (easy), instead of the current way of `airport_get().getRunway(rwy).getEnd(rwy).name` (cumbersome). I don't think I'm articulating it very well, but I promise on my life this is a huge improvement in code clarity and utility, and makes it very intuitive to work with.

---

Anyway, THE POINT here is to improve the separation logic! Once I got all the above stuff worked out, I added some "runway metadata" (a platform to put stuff like runway relationships-- how far they are all from each other, what is parallel/intersecting who, etc). Added into there what I needed to apply the rules from the 7110.65, para. 5-9-6, regarding parallel simultaneous dependent precision approaches. Woo! :smiley: So now, the game will dynamically apply the appropriate reduced separation minima to aircraft fully established on the localizer based on the different categories for how far away the runway centerlines are from each other, etc.

Tested the ever-living bejesus out of this, but there may still be some minor bugs around it, so testing is appreciated. Up next on this topic will be wake turbulence separation rules, especially now that the hard work is done!

Test: [erikquinn.github.io/atc](erikquinn.github.io/atc)


erikquinn:
THIS CODE HAS NOT BEEN MERGED YET (even though it says it has).

This was accidentally grouped in with another PR that was being merged, and due to my sloppiness, they were merged together. Rolled back that merge, re-did the one that was getting merged, and created a new PR for this code. Surprisingly, GitHub is not smart enough to see that this code is no longer merged (it was merged for about 15 minutes before I undid it).

PLEASE SEE #426 FOR THE ACTUAL STATUS OF THIS PULL REQUEST.

PR #432, opened 2016-04-08
Datablock Improvements

### LINKED ISSUE #28: Move aircraft tags

erikquinn:
Resolves #28.

There have been proposals for "draggable" data blocks, which would be nice, but doesn't seem entirely necessary to me to go that far, especially considering that this actually isn't even an option on almost any ATC scopes, nor is it particularly necessary. The issue here lies in the fact that the data blocks are swollen in size such that they take up a lot of space for the information they contain. Real scopes have them more compact, and that is part of what I have done here. All text is the same size, but wrapped in a smaller package, and one that will resize to fit the length of the callsign. With the smaller size, the _need_ to move datablocks is _drastically_ reduced because they simply don't overlap as often.

In addition (the primary purpose that led me on this noble quest), I added a function, `runMoveDataBlock(dir)` that a user can utilize with its shortkey, the tilde ( ` ), followed by an intercardinal direction on the number pad, in reference to 5 being the aircraft position. This is excatly how it is done in the real world; in fact, real STARS keyboards have little arrows on the numbers indicating this inherent functionality:

image
(Note the implementation is actually upside-down from this, to conform with normal-people-keyboards' numpad layout)

---

Here are some screenshots of how they look. Notice how very high levels of zoom are no longer necessary to achieve clarity in the final corridor. :+1: The aircraft on final in the screenshots are at absolute minimum spacing, so this visual clarity is "as bad as it gets".

overview

middlezoom

arrivalzoom

Demo: [3-minute demo video](https://youtu.be/UrhSiB46jkk)
Test: [erikquinn.github.io/atc/b/datablockImprovements](http://erikquinn.github.io/atc/b/datablockImprovements)

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -662,37 +662,36 @@ var Aircraft=Fiber.extend(function() {
   return {
     init: function(options) {
       if(!options) options={};
-      this.eid         = prop.aircraft.list.length;  // entity ID
-      this.position    = [0, 0];
-
-      this.model       = null;
-
-      this.airline     = "";
-      this.callsign    = "";
-
-      this.heading     = 0;
-      this.altitude    = 0;
-      this.speed       = 0;
-      this.groundSpeed = 0;
-      this.groundTrack = 0;
-      this.ds          = 0;
-
-      this.takeoffTime = 0;
-
-      // Distance laterally from the approach path
-      this.approachOffset = 0;
-      // Distance longitudinally from the threshold
-      this.approachDistance = 0;
-
-      this.radial      = 0;
-      this.distance    = 0;
-      this.destination = null;
-
-      this.trend       = 0;
-
-      this.history     = [];
-
-      this.restricted = {list: []};
+      this.eid          = prop.aircraft.list.length;  // entity ID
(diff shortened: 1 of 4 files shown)


headphase:
Looks nice!  I don't understand what you mean with:

> I added a function, runMoveDataBlock(dir) that a user can utilize with its shortkey, the tilde ( ` ), followed by an intercardinal direction on the number pad, in reference to 5 being the aircraft position. This is excatly how it is done in the real world; in fact, real STARS keyboards have little arrows on the numbers indicating this inherent functionality:

Are you saying that pressing one of the arrow keys moves all blocks on screen to that position relative to their respective aircraft? Or will it just move a mouse-selected block?

erikquinn:
@headphase It will only change the selected aircraft's fdb position. So if two are overlapping, click an aircraft and hit tilde, direction, enter. Example command: `AAL123 `6` would move AAL123's data block to the right side of his dot.

Specifying the position for all aircraft would likely be done through a menu setting, and not a command directly. On ATC scopes using STARS, this is the "LDR DIR" button within the DCB menu.
image

headphase:
Very cool; I'll look forward to this being merged :)

PR #434, opened 2016-04-09
Updated KJFK with some departures and better arrival traffic

jpeterson1346:


erikquinn:
@jpeterson1346 Thanks for the work here.... I was hoping somebody would get around to JFK! :smiley: 

I am seeing a few strange things as I look it over though. I see in commit 6cc63b2, you removed `afr.json`, `cca.json`, and `uae.json`-- was that intentional? They are probably used by other airports, and should stay in place.

jpeterson1346:
I think this was caused by an update from a "master" branch

erikquinn:
@jpeterson1346 ok. I was able to resolve the conflicts for you, but the merged code includes some alterations from what you had here, such as changing `"aircraft"` to `"fleets"` in your new airlines, so they are detected correctly by the parser (else, we get errors crashing the game), and correcting the format of the STARs you had added so they are understood by the code. If you like, you can have a look at the changes effected into the repository at commit 5c91006318925383c8a861d9c3fdd1f6a520a72e.

erikquinn:
**This code was merged in 455011d.**

Closing.

PR #437, opened 2016-04-11
New Airport: EDDT (Berlin Tegel)

Fechulo:
Berlin Tegel Airport is the main international airport of Berlin as well as the 4th busiest airport in Germany. Even though it's quite a small airport It receives a lot of traffic from all over Europe.
I hope you enjoy it!


erikquinn:
Looks good! Even though it hurts my brain a little... lol

PR #443, opened 2016-04-13
Holding patterns

### LINKED ISSUE #373: Racetrack Hold Pattern Instead of Circling
Planes in real life holds in [racetrack patters](https://en.wikipedia.org/wiki/Holding_%28aeronautics%29), not circles. Therefore it would be really nice to rewrite the `hold` command to enable this pattern. A simple implementation could be:
1. Player inputs `hold x` (`x` is either left or right);
2. Plane immediately turns `x` towards (360 minus the plane's current heading);
3. After the reciprocal heading is established, count one minute (or 10 nautical miles, which is a real life measure);
4. Plane turns `x` towards the original heading;
5. Count one minute or 10 NM.

erikquinn:
Resolves #373.

image

Implements proper holding patterns to replace the continuous-turn circles. Aircraft may now be told to proceed direct to a fix and hold there, along their inbound course. Alternatively, the same old same old also works, so you can tell them to `hold`, and they'll hold over their present position. If you like, you can also specify the turn direction with either `left` or `right` (default), as well as the leg length (only expressable in _minutes_ at this time), formatted like `1min` (default) or `3min`.

Examples:
- `hold`
- `hold left`
- `hold right`1min` (redundant, same as first)
- `hold left faith`
- `hold swels 3min` (long legs)

Note that this is not done entirely "properly"... they just make repeating 180 degree turns in the specified direction, and do not correct for wind. However, in anything other than a gale-force storm, I've observed no appreciable difference, and it seems to do the job well enough that I don't see a reason to spend time to get it to crab into the wind during the hold, though this is what a real pilot would have to do. But it looks a heck of a lot better than what we've got now! :smile:

Test: [erikquinn.github.io/atc](http://erikquinn.github.io/atc)

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -370,9 +370,6 @@ zlsa.atc.AircraftFlightManagementSystem = Fiber.extend(function() {
         this.fp.route.push("KDBG",airport_get().icao);
     },
 
-    parse: function(options) {
-    },
-
     /*
      Return this fms's parent aircraft */
     my_aircraft: function() {
@@ -396,6 +393,7 @@ zlsa.atc.AircraftFlightManagementSystem = Fiber.extend(function() {
         expedite: false,
         speed:    null,
         runway:   null,
+        legLength: null,
         fixRestrictions: {
           alt: null,
           spd: null
@@ -856,7 +854,6 @@ var Aircraft=Fiber.extend(function() {
         });
       }
     },
-
     cleanup: function() {
       this.html.remove();
     },
@@ -1027,8 +1024,7 @@ var Aircraft=Fiber.extend(function() {
           synonyms: ['t', 'h', 'turn']},
 
         hold: {
-          func: 'runHold',
-          synonyms: ['circle']},
+          func: 'runHold'},
 
         land: {
           func: 'runLanding',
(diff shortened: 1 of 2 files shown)


headphase:
sexy!

erikquinn:
@headphase lol I thought so too!

Alpi-no:
Is there a chance to add a way to make incoming planes automatically hold at the first waypoint they have been assigned for from the center?

erikquinn:
@jakobeng1303 No, you would have to tell each one to hold at the fix once they enter your airspace. That could be added, yes, but I don't know of anywhere that actually does this, so I would be hesitant to implement it if it isn't done in the real world.

Alpi-no:
@erikquinn I'm not 100% sure and as I don't know any ATCs I can't really tell, but I thought if you haven't been cleared for a star, you must not proceed further than the holding before your star. At least in Vienna, my home airport, I regularly see planes in holdings before the star. I know the atc could have told them to stay there, but I'm quite unsure about this.

erikquinn:
@jakobeng1303 yeah I see where you're coming from, and there's some truth to it. Basically, the aircraft will get an IFR clearance from the departure airport to the arrival airport, including any SIDs/STARs. During the flight, if the arrival airport gets too busy, some sort of macro-level Traffic Management Unit will start issuing holds to reduce the arrival rate to something the arrival airport can safely handle. So they START with the full clearance including the STAR, and as needed, ATC will tell inbound aircraft to hold, and usually at a waypoint along the STAR that has a published holding pattern. That's why you're likely to see them holding in the same place when they hold, but that would only happen during especially busy times, as a good traffic management crew will be able to avoid the delays in holding by delaying each departure a few minutes at the very beginning. I look forward to one day learning more about TMU ops; it's fascinating how they work! :smiley:

Alpi-no:
@erikquinn oh oops :D I originally thought that you'd be cleared from the holding and not right from the start!

erikquinn:
@jakobeng1303 Nope :smile: They will actually amend the clearance if you need to hold, changing the clearance limit to the place where they're holding, and issue hold instructions. So you're right that the clearance limit is changed, but they start with the whole thing, and then the clearance limit is moved backwards along the route from the destination airport to some fix along their route as needed.

### Example:

Initial IFR clearance along the full route, given while on the ground at Boston-Logan:

> "N123, cleared to the San Francisco airport via the Hyland Four departure, HYLND transition, then as filed. Climb and maintain 3,000, expect Flight Level 360 10 minutes after departure, departure frequency 127.87, squawk 6132."

Then later on, as SFO weather turns to crap, Oakland Center would amend the clearance:

> "N123, cleared to the FLOWZ intersection. Cross 20 miles east FLOWZ at and maintain FL210, hold east, as published, expect further clearance 0315, due to low ceilings and deteriorating visibility in the bay area."

image

jakobeng1303:
@erikquinn But if you think of it that way, why is the tour clearing planes and telling them to taxi? At least in LOWW thats definitely not the case, nor do I think in many other airports ;) :D

erikquinn:
I'm not sure I understand what you mean?

Alpi-no:
@erikquinn forget about it, if it will be important later on, I'll explain it :D

Fechulo:
@erikquinn I think, I know what @jakobeng1303 is talking about. In most (if not all) airports in Europe, the initial clearance includes the SID and route, **NOT** the STAR. Once the aircraft gets a bit closer to the arrival airport, it is cleared to a STAR. While on the STAR, the aircraft can also be given clearance for the final approach or transition, if not it must hold at the last waypoint of the STAR.

Here's two examples:
Heathrow
egll example

Munich
eddm example

Alpi-no:
@Fechulo just to confirm, that is the proof, of what I thought about earlier, isn't it? I'm actually quite surprised about it, but it kinda makes sense :D

Fechulo:
> I thought if you haven't been cleared for a star, you must not proceed further than the holding before your star

I'm trying to prove this point :)

Alpi-no:
Thanks :D

erikquinn:
Huh... I had no idea they did it like that some places! The more you know, I suppose... :smiley:

Well that should be easy enough to emulate if we wanted to... could just add a holding waypoint within the STAR. Might get messy in the sim though, for a few reasons. Regardless, that's very interesting. :+1:

Issue #444, opened 2016-04-13
is ATC broken?


bolero3232:
All I get is the green screen with the circles. No runways, no sids, no names... nothing. Mouse has no effect, no zoom, no move.

Fechulo:
In which airport are you experiencing this problem?

Fechulo:
@erikquinn Seems to be a problem with AUA not loading. Could it be have to do with the changes in #431 ?

bolero3232:
All airports.
What's #431?

bolero3232:
#431 got it

Fechulo:
@bolero3232 It's actually not all airports, the problem is that when one airport fails, all airports after it fail too. The problem seems to be in the airports which include Austrian Airlines, which leads me to believe that this is somehow the cause. Since I'm not the expert here, I am not 100% sure.

bolero3232:
So what do I do about it?

headphase:
Testing now and at first I thought it might be due to the uncapitalized "b77l" reference in the fleets of those airlines updated by [commit](https://github.com/zlsa/atc/commit/57dec34acc0c077d25fe9a121b573d0255e2602b) but after recapitalization I'm still having issues with random airports.

Working:
KJFK 
KMSP
KSFO
KLAX
KLAX90
EBBR
EDDM
EHAM
ENGM
LTBA
SAEZ
SBGL
SBGR
UUDD
VHHH

Problems occuring / Loading Failing intermittently:
EDDH
EDDT
EIDW
LOWW

Not loading at all (?)
EGLC
EINN

erikquinn:
Yeah it's apparently actually due to about a dozen airlines not being called to load. It was like 15 or so, meaning the loading issues would be widespread across the airports, since they all are likely to use one of the affected airlines. Seems impossible that I missed this, as I didn't come across it while testing, but somehow it made it out to the live site. Sorry, this is 100% on me. :disappointed:

I've got them all added in now on a local branch and am just testing the (ahem) _stuff_ out of it before pushing in a hotfix.

headphase:
Really? That's odd I was looking through airlines.js on the live site and it seems every company is included on the list 👀

erikquinn:
@headphase Just not some of the new ones from #431 apparently. Four are good, the rest are never loaded in `airlines.js`. I was equally surprised not to have noticed it...

@all
**Could some of you verify it is now fixed at this link please?**
[...]

PR #453, opened 2016-04-15
Fix IE canvas compatibility issue

### LINKED ISSUE #450: Internet Explorer Compatibility
Not sure for how long, but it appears the game is no longer drawing properly in Internet Explorer. Other than that, nothing seems to be wrong. No errors logged in console. Reported in #444.

image

erikquinn:
Resolves #450.

IE doesn't natively support String.fromCodePoint(). Thus, it had to be manually added to String.prototype to get IE to 'cut the crap and do [its] job'. Implemented this by checking to see if the function is defined, and if not, defining it. This way, the game can never be broken by the lack of support for this function, regardless of browser.

### DIFF
--- a/assets/scripts/modules.js
+++ b/assets/scripts/modules.js
@@ -74,6 +74,31 @@ var RELEASE=false;
 
 //////////////////////////////////////////////////////////////////////////////////////////
 
+/********* Various fixes for browser issues *********/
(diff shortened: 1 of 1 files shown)


erikquinn:
Verified by the reporter in #444 that this issue was resolved by these changes (prior to merge).

image

Issue #438, opened 2016-04-11
Best practices for decluttering canvas when an airport has many SIDs?  Suggestions welcome


headphase:
Hey guys looking for some brainstorming help; I'm working on KBOS and it turns out there are a LOT of SID waypoints due to the extensive noise abatement procedures and obscene number of runways at an airport of its size haha 😂

So far I have coded 3 out of 9 SIDs and the canvas looks like this:

![kbos](https://cloud.githubusercontent.com/assets/6735197/14437775/db4d6d14-fff0-11e5-9336-8872f4c5b256.png)

Pretty unacceptable imho!  So what is the best way to balance player experience with detail?  So far I'm thinking of:
- Hiding the names of waypoints which are **not** shared by multiple runways.
- Eliminating the SID path markers (I don't really want to, but I'm not sure what other option there is to keep things relatively clean?

Any other ideas?

erikquinn:
Oh please, we all know Vienna is the ultimate winner of the "beautiful nightmare" award!

![image](https://cloud.githubusercontent.com/assets/5103735/14438291/f0b69276-ffea-11e5-805b-824c5694f06b.png)

:laughing: No, but seriously, we do need to figure out how to deal with this, because I agree, it's so cluttered that any possible benefit is lost. Airports with complex SIDs are usually that way because the SIDs make it EASIER on the controller, and by shoving all of them in the user's face, we're kind of doing a little of the opposite. To be honest, when real video maps look as crazy as they do (see #423), I don't really object to a big old pile of fixes. But I don't personally see a reason to have the blue lines. That said, I usually just turn them off by clicking the little taking-off-airplane in the bottom menu-- perhaps that could default to off, if a few people can come to a consensus? That could be a solution, but would make it tougher on new players to predict who is going to go where... food for thought, anyway

Fechulo:
@erikquinn LOWW was so crowded, that I had to remove most of the already existing fixes! Regarding #423 I don't really know what any of those lines and shapes mean. If they serve the same purpose as the blue SID lines, I think we should definitely replace them. Another option is to hide the blue lines for the unused runways or we could let the player decide which ones should be displayed and which ones shouldn't.

@headphase KBOS is looking really good so far. Great job!

erikquinn:
@Fechulo I can only imagine how much work that must have taken... there is a _lot_ going on there. :+1: 

[...]

PR #455, opened 2016-04-18
Update EDDM with new SID format

Fechulo:
I should be known as "the one who clutters everything".

... sorry :(

### DIFF
--- a/assets/airports/eddm.json
+++ b/assets/airports/eddm.json
@@ -2,9 +2,9 @@
   "name": "Franz Josef Strauß International Airport &#9983",
   "level": "beginner",
   "radio": {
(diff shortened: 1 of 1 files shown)


erikquinn:
Looks good! And not at all; _realism_ is the goal, and you've done a lot on that end, which has made gameplay much better :+1: it's really an issue of how much of that complexity we should be displaying, which is something that needs to be fixed soon. But please keep doing what you're doing, because this is **really good stuff**! :smiley:

Fechulo:
Thanks! I'm currently working on ENGM, which I might finish today. I think that the solutions proposed in #438 should definitely fix this cluttering issue.

Issue #439, opened 2016-04-12
Instrument Procedures with Vectors & Holds


jpeterson1346:
Is there a way to incorporate radar vectors between a departure's climb and the departure itself (e.g., DEEZZ4 departure at KJFK:

"TAKEOFF RWYS 22L/R: Climb heading 225° or as assigned by ATC, thence . . . .

. . . . expect vectors to DEEZZ, then on track 296° to HEERO. Maintain 5000. Expect clearance
to filed altitude/flight level within ten (10) minutes after departure."

instead of having a departure fly from say "YNKEE" directly to DEEZZ and then continue as filed?

ref: [http://flightaware.com/resources/airport/KJFK/DP/all/pdf](url)

erikquinn:
Currently not, but since this is such common practice, it would definitely be a good addition to the SID functionality.

jpeterson1346:
Most definitely agree; as it stands now, I cannot issue vectors after an initial climb out of JFK through the JFK3 departure (#440). Thus, the JFK3 departure is unusable at this time.

erikquinn:
See #651 (closed as dupe).

erikquinn:
The ATC repository is being migrated to it's new home at [https://github.com/openscope/openscope](https://github.com/openscope/openscope), 
and thus, all issues are being closed. If this is still an issue with the latest version of the sim 
(accessible at [http://www.openscope.co](http://www.openscope.co)), or is a feature you still think we are lacking, 
please reopen the issue at the new repo.

*Please note that the vast majority of these issues have been copied to the new repository, or else are covered by other issues created there. See the below screenshot for what it looks like when your issue is known in the new repo:*

![image](https://cloud.githubusercontent.com/assets/5103735/22630816/5a32d794-ebce-11e6-9c46-d61a262e2b23.png)


 Thank you!


Closing this issue.

Issue #407, opened 2016-02-27
Previewing/Showing list of assigned fixes in strip


kevini15:
I've seen a picture of it somewhere in the Issues section already, but the main topic of that was something else... Wouldn't that be awesome, if there was like a list of fixes the aircraft has been assigned to?

Thanks

erikquinn:
The ATC repository is being migrated to it's new home at [https://github.com/openscope/openscope](https://github.com/openscope/openscope), 
and thus, all issues are being closed. If this is still an issue with the latest version of the sim 
(accessible at [http://www.openscope.co](http://www.openscope.co)), or is a feature you still think we are lacking, 
please reopen the issue at the new repo.

*Please note that the vast majority of these issues have been copied to the new repository, or else are covered by other issues created there. See the below screenshot for what it looks like when your issue is known in the new repo:*

![image](https://cloud.githubusercontent.com/assets/5103735/22630816/5a32d794-ebce-11e6-9c46-d61a262e2b23.png)


 Thank you!


Closing this issue.

Issue #379, opened 2016-02-07
Issuing a "heading" command after "fix" (vice versa)?


siyuanr:
Currently `heading` and `fix` commands would override each other. Is it possible to implement a consecutive command structure here via a `then` clause? For example, `f ATENA then h 360` would order the plane to turn 360 after reaching `ATENA`.

Or we could use a `cross` clause to note the specific point of executing the next command. For example: `h 360 cross ATENA h 340` tells the plane to head 360 for now; head 340 after `ATENA`. We could then have:
- `cross ATENA`: the "if" clause is the plane reaching a fix;
- `cross ATENA XXX`: the "if" clause is the angle of line between plane and fix reaching XXX (crossing a radial);
- `cross ATENA XX` or `cross ATENA X`: the "if" clause is the distance in NM between plane and fix reaching XX or X (crossing a DME measurement).

The top use case for consecutive commands is SID. It may also come in handy for ILS approaches.

sixeyeco:
@siyuanr :+1: 

> It may also come in handy for ILS approaches. 

I'm really challenging myself with [the "reopened airport scenario" on my fork](sixeyeco.github.io/atc/). Typically, I issue `f faith wetor`. Then, I have to issue another command to land. The ability to issue `f faith wetor then L28L` would be incredibly beneficial.

tedrek:
This topic is currently being discussed at #315

erikquinn:
Duplicate of #315. Closing.

Issue #315, opened 2016-01-01
Further supported behaviors for waypoints


tedrek:
From @erikquinn  in #313: Can we assign crossing restrictions for a fix/waypoint.  (eg. "descend at pilots discretion, cross BERRA at and maintain 1-1 thousand at 2-5-0 knots")? this is common in enroute and STARs, and would be both cool and helpful to emulate.

tedrek:
After #313 is merged, it should certainly be easier and is one of the things I had envisioned when writing the feature.

Right now, the system supports setting an altitude and speed until a given waypoint followed by a different altitude and speed until the next waypoint.  To do crossing restrictions we'll need to add an altitude/speed by the given waypoint and the associated projection to start the descent/deceleration.

Another thing which needs to be implemented is exposing this functionality to the controller so you can assign some clearance in relation to a waypoint.  I had been thinking something like adding three verbs: 'until', 'by', and 'after'.  Example uses would be 'BAW831 by SENZY speed 180 altitude 2000', 'BAW831 until WAMMY speed 210 altitude 8000', 'BAW831 after WAMMY speed 250 altitude 10000'.

The question I have is, what happens when an aircraft is on a route where speed and altitude are assigned for multiple waypoints and the controller assigns a plain speed or altitude.  My expectation is that speed and altitude will be followed through subsequent waypoints unless the controller amends it but I would really like to hear what some other folks would expect.

tedrek:
The other feature which I wanted to implement using this base was holding patterns followed by automatic holds on STARs if aircraft have not been cleared to further fixes.

Maverick283:
If I get you right, you're asking what happens when a aircraft was told "descent altitude 5000 by ANYFX " (I believe this is the actual syntax @ericquinn?) and is told "maintain altitude 7000" before it reaches ANYFX or 5000 or 7000?
In this case the pilot would ignore the altitude part of the first instruction, and maintain 7000 but keep flying to the fix. Basically if one instruction overrides another, the more current one will be executed...

erikquinn:
@tedrek in place of until/by/after, could I suggest until/cross/depart? that would be more in line with phraseology of the commands:

> > "AAL482, maintain 190 knots to FINKA"
> > "UAL3185, cross LENDY at Flight Level 190 at 250 knots"
> > "KAP515, depart BOWAN heading 360, vector for visual runway 28 approach"

[...]

PR #464, opened 2016-04-30
Airways, Legs, Rerouting, & So Much More

### LINKED ISSUE #350: Taxi/Wait puts flights on the runway causing incursions/collisions.
At SFO there was a collision between a flight was cleared to take off and one that was waiting at the same runway. 

The only command I issued to the latter flight was 

> C 100 SP 450 W 28R 

https://www.youtube.com/watch?v=6xovvGCUnVw

I'm certain it would help if I mention that I was playing a forked version that's located here: http://sixeyeco.github.io/atc/ but the only edit I've made was to delete the first way point for SID Offshore.

### LINKED ISSUE #430: Aircraft taxiing to runways not approved by SID
I was reading through issue #324 and thought this was handled by the complex SIDs update, so maybe I'm coding something incorrectly but the image below should explain it; aircraft who have filed the SKORR3 departure should depart on rwy 31L/31R according to the chart (and I've omitted the other runways in the code) however pilots are still taxing to other runways when given the unspecific "taxi" clearance.  I thought that the game did not allow aircraft to use a SID if it was not available for the default runway assignment?

skorr

### LINKED ISSUE #461: Airways
Adding the ability for aircraft to follow airways while in flight could add a lot, depending on the airspace. Particularly with respect to eventually adding an En Route mode to the simulator, this would be critical to complete before that could happen. Some reworking of the fms would be needed, but if done correctly, the implementation and addition of procedures like SIDs/STARs/IAPs, airways, published holding patterns, hybrid RNAV SIDs (where vectored for part, then cleared direct to join the departure) could be extremely straightforward. The key would be to change the perspective of the fms from waypoint-based to leg-based. Eg, currently, it sees:

`KSFO --> SENZY --> WAMMY --> SEGUL --> CYPRS --> MCKEY --> DAISY --> BENET --> SXC --> AVOLS --> PACIF --> HUBRD --> OCN --> VISTA --> ROBNN --> BONDO --> JLI --> KUMBA --> NICKK --> IPL --> etc`

This route is really a reasonable one, and in reality would be a much longer list of fixes. To clear it up, routes are written in reference to a joining point, a procedure/airway, and an exit point. This is how we would implement legs, and the same route in that notation would be like the below:

`KSFO.OFFSH9.SXC --> SXC.V458.IPL --> etc`

This way, by looking at the LEGS, we can easily edit, move the legs around, add a fix between legs, etc, whereas the same operation on a long list of fixes would be difficult, because you simply wouldn't know where a procedure started or ended.

The point is that airways would be a great addition, and I'm working on throwing something together now. Any feedback or ideas are welcome.

erikquinn:
This is finally pretty much ready to go!

Includes a lot of back-end work to improve the way the fms handles route logic, so that routes can be manipulated in a way that is similar to real flight management systems, thus making it easier to achieve more complex routing operations, moving things around, and just overall enabling us to achieve a higher level of capability.

Basically, here's the jist of what happened to the fms:

> The current fms is set up to navigate between "waypoints", which are manually created objects, not associated with a particular class, that are made in the fms's `addWaypoint()` member function. You can add a whole bunch of these waypoints to the fms, and it will tell the aircraft to aim to each waypoint. As it passes each one, that waypoint is removed from the fms, and the others shift over, and the a/c goes to the next one. The process repeats until no waypoints remain, and then the aircraft will pass through the last waypoint and maintain its altitude/speed/heading when it passed its last waypoint.  
> In this proposed pull request, the fms has a modifiable flight plan route, which has corresponding "legs" that tell the fms where to steer the airplane. Each leg contains a series of waypoints. The purpose of doing it this way is to condense and clearly group together waypoints when they are part of an instrument procedure; for example, a SID or STAR will be maybe 5-10 fixes long, and may have specific altitude/speed limitations at each fix. By having these 5-10 waypoints contained within a `Leg`, you can clearly differentiate these fixes from the rest in the fms. If you wish to change to a different SID (perhaps because the one the aircraft had filed for is invalid from the runway you've taxiied them to), and all the waypoints were just loose in the fms, you wouldn't know which ones (or how many) to remove. Here, you just find the route that has `route.type == "sid"`, and replace it with a new leg, as simply as saying `fms.insertLegHere(new zlsa.atc.Leg({type: "sid", route: "KSFO.MOLEN7.ENI"}));`. Boom. The existing logic builds EVERYTHING else for you, as long as `MOLEN7` is a valid SID at `KSFO`, and `ENI` is a valid transition of that SID. If not, it'll throw an error and let you know the route is invalid. This is a better way to handle is because you can add/remove things easier, and better recognize when an aircraft is able to climb/descend via a SID/STAR, or whether they are on an airway, etc, which can be useful in many other applications.

_In Reference to Active Issues:_
- Resolves #350, via making a/c on ground invincible to all collisions (temporary solution)
- Resolves #461, by adding capability to parse, understand, use, and manipulate routes to accommodate airways
- Resolves #430, by moving the departure runway member variable from `fms` to `Aircraft`
- Enables us to easily add a fix for #439.
- Enables us to easily add a fix for #379.
- Enables us to easily add a fix for #407, and partially addresses the topics discussed by adding a `sr` command, for "say route"

Technical Changes:
- Added class `zlsa.atc.Waypoint`, based on old waypoint format
- Added class `zlsa.atc.Leg`, which is a functional part of the flightplan route, and contains a series of Waypoint objects
- Set up straightforward framework for stuff like vector/RNAV hybrid SIDs (#439), departing/crossing fixes on certain headings or at certain speeds (#379, #315)
- Fixed math error with `offset` calculations (used to be member fxn of `Runway`. now in `util.js` and set up to be more abstracted and versatile (this could help runway intercept weirdness and lays groundwork for intercepting radials or airways as if they were ILS localizers)

_Notable Features/Changes Included:_
- `route`, `rr` ("reroute"), and `sr` ("say route") commands
- Taxi time set to 3 seconds, and aircraft in the queue no longer are waiting around for days if you taxiied a bunch back to back. As long as each has had their 3 seconds, they will appear on the runway as soon as the departure ahead of them takes off. Note: All conflicts/crashes disabled while on ground (temporary fix).
- removed `proceed` command, as it is made moot by introduction of the more powerful and real-world accurate methods of rerouting aircraft via `route` or `rr` (reroute) commands

---

I have tested it pretty thoroughly, and continue finding small things that have broken, and I am sure there are more, but nothing that will hinder the gameplay, and nothing that can't be very easily fixed... It's just tough to track down them all, and I think I have just about everything running properly. There may be issues with something like rerouting, then assigning a fix, then clearing for an approach, then adding a hold (or something crazy like that that I didn't test), but I'm sure we will come across any such issues eventually if they exist, and we can deal with them then. This should be a really big step forward though, and enable some _really cool stuff_.

Testing on this is appreciated; let me know if you find anything broken!

Airways available in demo (not included in this PR) include:
`"V87": ["POPES", "SGD", "REBAS", "SFO", "OSI", "SANTY", "MOVER", "SNS"]`
`"V107": ["BOARS", "PYE", "MICRA", "COMMO", "OAK", "DECOT", "IMPLY", "MISON", "MABRY", "VINCO", "CATHE", "PXN"]`

Note: Aircraft can be made to follow airways via the `route` command, which adds a new Leg. Format your input with single-dots for airways, or double-dots for fixes. Alternatively, do `rr` instead of `route` to replace the aircraft's route entirely with a route of your own making (same format). An example of proper route format is `KSFO.OFFSH9.SXC.V458.IPL.J2.JCT..LLO..ACT..KACT`

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -566,11 +566,12 @@ zlsa.atc.AircraftFlightManagementSystem = Fiber.extend(function() {
       if(prev && !curr.speed) curr.speed = prev.speed;
     },
 
-    /** Insert a waypoint before the *curent* waypoint
+    /** Insert a waypoint at current position and immediately activate it
      */
-    prependWaypoint: function(data) {
+    insertWaypointHere: function(data) {
       var prev = this.currentWaypoint();
       this.currentLeg().waypoints.splice(this.current[1], 0, new zlsa.atc.Waypoint(data));
+      this.update_fp_route();
 
       // Verify altitude & speed not null
       var curr = this.currentWaypoint();
@@ -587,6 +588,9 @@ zlsa.atc.AircraftFlightManagementSystem = Fiber.extend(function() {
       this.legs.splice(data.firstIndex, 0, new zlsa.atc.Leg(data, this));
       this.update_fp_route();
 
+      // Adjust 'current'
+      if(this.current[0] >= data.firstIndex) this.current[1] = 0;
+
       // Verify altitude & speed not null
       var curr = this.currentWaypoint();
       if(prev && !curr.altitude) curr.altitude = prev.altitude;
@@ -598,6 +602,7 @@ zlsa.atc.AircraftFlightManagementSystem = Fiber.extend(function() {
     insertLegHere: function(data) {
       data.firstIndex = this.current[0];  // index of current leg
       this.insertLeg(data); // put new Leg at current position
+      this.current[1] = 0;  // start at first wp in this new leg
     },
 
     /** Insert a Leg at the end of the flightplan
@@ -611,6 +616,7 @@ zlsa.atc.AircraftFlightManagementSystem = Fiber.extend(function() {
      */
     appendWaypoint: function(data) {
       this.currentLeg().waypoints.splice(this.current[1]+1, 0, new zlsa.atc.Waypoint(data));
+      this.update_fp_route();
     },
 
     /** Switch to the next waypoint
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -1,8 +1,9 @@
+/** Details about aircraft in close proximity in relation to 'the rules'
+ */
 zlsa.atc.Conflict = Fiber.extend(function() {
   return {
     init: function(first, second) {
       this.aircraft = [first, second];
-
       this.distance = vlen(vsub(first.position, second.position));
       this.distance_delta = 0;
       this.altitude = abs(first.altitude - second.altitude);
@@ -97,6 +98,7 @@ zlsa.atc.Conflict = Fiber.extend(function() {
      * Check for collision
      */
     checkCollision: function() {
+      if(this.aircraft[0].isLanded() || this.aircraft[1].isLanded()) return;  // TEMPORARY FIX FOR CRASHES BTWN ARRIVALS AND TAXIIED A/C
       // Collide within 160 feet
       if (((this.distance < 0.05) && (this.altitude < 160)) &&
           (this.aircraft[0].isVisible() && this.aircraft[1].isVisible()))
@@ -142,11 +144,11 @@ zlsa.atc.Conflict = Fiber.extend(function() {
 
       // Check for the same runway, different ends and under about 6 miles
       if ((!this.aircraft[0].isTaxiing() && !this.aircraft[1].isTaxiing()) &&
-          (this.aircraft[0].fms.currentWaypoint().runway != null) &&
-          (this.aircraft[0].fms.currentWaypoint().runway !=
-           this.aircraft[1].fms.currentWaypoint().runway) &&
-          (airport.getRunway(this.aircraft[1].fms.currentWaypoint().runway) ===
-           airport.getRunway(this.aircraft[0].fms.currentWaypoint().runway)) &&
+          (this.aircraft[0].rwy_dep != null) &&
+          (this.aircraft[0].rwy_dep !=
+           this.aircraft[1].rwy_dep) &&
+          (airport.getRunway(this.aircraft[1].rwy_dep) ===
+           airport.getRunway(this.aircraft[0].rwy_dep)) &&
           (this.distance < 10))
       {
         if (!this.conflicts.runwayCollision) {
@@ -185,8 +187,8 @@ zlsa.atc.Conflict = Fiber.extend(function() {
 
       // Established on precision guided approaches
       if ( (a1.isPrecisionGuided() && a2.isPrecisionGuided()) &&
-           (a1.fms.currentWaypoint().runway != a2.fms.currentWaypoint().runway)) { // both are following different instrument approaches
-        var runwayRelationship = airport_get().metadata.rwy[a1.fms.currentWaypoint().runway][a2.fms.currentWaypoint().runway];
+           (a1.rwy_dep != a2.rwy_dep)) { // both are following different instrument approaches
+        var runwayRelationship = airport_get().metadata.rwy[a1.rwy_dep][a2.rwy_dep];
         if (runwayRelationship.parallel) {
           // Determine applicable lateral separation minima for conducting
           // parallel simultaneous dependent approaches on these runways:
@@ -260,6 +262,8 @@ zlsa.atc.Conflict = Fiber.extend(function() {
   };
 });
 
+/** Definitions for characteristics of a particular aircraft type
+ */
 var Model=Fiber.extend(function() {
   return {
     init: function(options) {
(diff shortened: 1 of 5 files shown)


erikquinn:
Anybody have a chance to play around with this? When it is merged, it will open a bunch of doors for things that can be improved/fixed, but I don't want to merge it until there is a consensus and it is certain that it doesn't break anything for anyone. I've tested it thoroughly myself obviously, but still would like a confirmation, because it is quite a big change... I have a few items on my list of things to do next but work on them cannot begin until this goes in. Anybody willing to try to break it? :hammer: :fire: :computer:

Fechulo:
I'm a bit confused. At the moment, what is the use of airways for a player? I understand they will be extremely useful if we have a center control, but right now, the player should be happy by just routing the aircraft via the sid's. Am I missing anything? What exactly would you like me to test?

erikquinn:
Correct, it doesn't really make a difference at this stage, but in addition to adding capability for airways, it reworks the fms in a way that could break just about every single individual thing. I'm just looking for a few people to verify that whatever they usually do at their usual airport of choice has not been broken by anything in here. All I really need is to make sure it looks like NOTHING has changed to the user. No need to go in-depth with each function, as I've tested all of them thoroughly. It's just that when I get in that mindset, I'll rarely miss a small detail, but (believe it or not) I could easily overlook something like "when you clear somebody for an approach and then change the speed, the game crashes".

So I am just looking for a verification that nothing totally obvious is totally broken, because otherwise this change shouldn't make the slightest difference to the users.

Fechulo:
Thanks, I understand now. I'll definitely take a look at it and tell you if I find anything.

erikquinn:
@Fechulo Thank you!! :smile:

Fechulo:
First bug: In some airports (most noticeable in EGLL) the arrival aircraft spawned at the start of the session will try to go back to the first waypoint in the STAR, instead of flying directly to the closest waypoint. It can be easily seen if the aircraft projected path is set to show always.

Alpi-no:
So based on this can we program stars?

erikquinn:
@jakobeng1303 Not yet, this just makes it easy to add logic for stars. They will work with EXACTLY the same format as SIDs, so if you were getting anxious you could always have a look at a few SIDs, and get some STARs ready for your favorite airport(s). I suspect it will be 7 days or less until we can do stars AND hybrid RNAV SIDs (aka 'vector departures')

Alpi-no:
Nice! I will check out the data format and get some ready!

erikquinn:
@Fechulo is that the only issue you've encountered? Just FYI, I'm targeting tomorrow afternoon (~2000z) for merging this if no other bad stuff is found, so stuff like STARs can start getting some attention.

Fechulo:
I'll do some more testing now, but I haven't found anything else.

Fechulo:
I found another bug, Instead of explaining it, [here's](https://youtu.be/6rjzommeAr8) a video! (it should be HD in a few minutes)

erikquinn:
@Fechulo Thanks for the video, it's super helpful! In fact, I actually decided to follow suit to show the slightly different intended workflow that would avoid this issue, and it's one I wasn't seeing because I was playing differently... I'll have a look at the bug and get it fixed.

Here's an [explanation and demo](https://youtu.be/0QYn7aMG_dA) of the intended way to achieve the control instructions you were giving.

Fechulo:
Thanks, I wrongly assumed the aircraft wouldn't remember their original route so that's why I gave them the fix command. 

Regarding the 10 dme arch numbers, they are just the number in the 10 dme arch, not a heading. I basically paste the coordinates for the entire arch, number them from 1 to 200 (or however many coordinates there are) and then I delete the ones I don't need. Instead of renumbering them I just leave them like that, resulting in nonsensical numbers. This "behavior" of mine can also be seen in EDDT.

As soon as I get back home tomorrow (~1700z) I'll do some more testing. I'm really looking forward to this being merged and expanded.

erikquinn:
Fixed the issue with the `fix` command in 11cdb27. It apparently wasn't that there was no altitude at the previous waypoint, it was as simple as the `currentWaypoint()` function returning null because `fms.current` was not updated correctly after inserting a leg _before_ the current one, which resulted in an out-of-range error if the aircraft was at the (n+1)th waypoint of the original leg if the inserted leg had n waypoints or fewer.

Alpi-no:
@erikquinn can I add the stars now?

erikquinn:
@jakobeng1303 Not quite yet, the sim still doesn't know what a STAR is. This pull only changed some things about the way the fms thinks. Now that it is a little smarter, we can teach it to read STARs in a matter of a few hours work, which is forthcoming. I'll create an issue to track the progress on that.

Alpi-no:
@erikquinn Ah now I think I got it :D

PR #465, opened 2016-05-02
Updated sas airline

blt950:
Updated the name and callsign which were incorrect.
Source: https://airodyssey.net/reference/airlines/


erikquinn:
@blt950 Thank you! :+1:

PR #467, opened 2016-05-02
Use iteration for callsign generation - Fixes #466

### LINKED ISSUE #466: Maximum call stack size exceeded
Has to do with the logic for selection of callsigns. As the game plays for a long time, it "uses up" most of the valid flight numbers, and the way callsign selection seems to work is that is picks a random number in the range, checks if it hasn't already been used, and if it _has_ been used, it gets a new one, and repeats the process. There is a limit on how many loops deep you can go with this kind of thing (like maybe 100 or so?). Clearly, as more and more callsigns are used over time, we can exceed whatever that limit is, which will crash the game.

This will only happen after HOURS of playing (or using `timewarp` to speed up the time a few hundred times). Not a big deal, but might as well fix it sometime.

image

tedrek:
Fixes #466 

Use iteration rather then recursion for callsign generation.  There is no stop condition in case all callsigns are used so it's possible this would cause a freeze.

I think this also fixes a bug where duplicate callsigns were added to the list of callsigns.

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -2725,18 +2725,12 @@ function aircraft_generate_callsign(airline_name) {
 }
 
 function aircraft_callsign_new(airline) {
+  var callsign = null;
   var hit = false;
   while(true) {
-    var callsign = aircraft_generate_callsign(airline);
-    hit=false;
-    for(var i=0;i<prop.aircraft.callsigns.length;i++) {
-      if(prop.aircraft.callsigns[i] == callsign) {
-        callsign = aircraft_callsign_new(airline);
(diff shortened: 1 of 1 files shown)


erikquinn:
Works! :+1: It does eventually cause that freeze, but I added a line to remove the callsign from `prop.aircraft.callsigns` when an aircraft is removed, so that callsign is made available to subsequent new aircraft.

Combined, it eventually still gets gunked up, but presumably just from there being like 500 departures waiting on the ground for taxi. lol

On my system, it now runs for about 58 real-time hours before dying. Somehow I have a feeling that will be sufficient... :laughing:

PR #474, opened 2016-05-05
More fixes for runway 11

Alpi-no:
I added some more fixes for runway 11 to help players line up planes efficiently on runway 11. they could be very helpful for stars in the future!


Fechulo:
I believe those fixes were there in the first version of the airport, however after complex SID's were added, the airspace became really crowded, so I decided to remove them. But since possible solutions for the cluttering of airports are currently being discussed, I don't see why this shouldn't be merged.

Alpi-no:
An idea I just had was, that unless you have selected a departing aircraft all the fixes not used for arrivals are a little faded so you still see them, but they are not too prominent

Fechulo:
It has been discussed in #438 and various ideas came up. The best solution @erikquinn thought of is to either: 

> show available SIDs only valid from the runway the aircraft is currently using, on, or planning to use or
> show available SIDs only valid from the runway best aligned with the wind

With the recent addition of the radar video maps, came a button to hide all fixes which could also be an easy way of solving this issue.

erikquinn:
@fechulo actually the button only appears if an rvm is used... If we removed fixes on non video map displays, we'd be left with like nothing haha

And for the record, I now think the former of my ideas above is the better of the two... It should show the SIDs valid from the aircraft's `rwy_dep` when the aircraft is selected. Alternatively, when the SID button is turned on, they all show (it would default to being off).

Fechulo:
Yeah, making all fixes disappear would be the ugliest and laziest solution, but I guess it would "work". I think so too, the first one seems the most sensible, as some players might use a different runway than what the game thinks would be the correct one to use.

Alpi-no:
I'm sorry to ask this, but due to my lack of experience with Github I have no clue what to do now :D , am I supposed to co close this pull?

Fechulo:
I am actually just a very active contributor, so I don't have the ability to merge pull requests. You just have to wait until someone merges your PR and then your code will be in the game.
Feel free to ask anything you want to know!

Alpi-no:
thanks!

erikquinn:
If @Fechulo likes it, I like it! :wink: Here we goooo!

PR #478, opened 2016-05-06
Add polygonal boundaries at EDDF

Fechulo:
This was more difficult than I thought it would be, since the airspace is made up of several areas which change depending on the runways in use:
airspace


erikquinn:
Tested this, #480, and #481 together. All look good. Merging.

PR #480, opened 2016-05-08
Add Polygonal Boundaries at EDDH

Fechulo:
This PR adds polygonal boundaries to Hamburg, it improves the SID's and adds a few more airlines for realism. When I first updated EDDH, I was not able to find a reliable source of coordinates for the fixes. However, while working on EDDF, I found one! The changes to the SID's are mainly visual, as you can see in the pictures:

Before:
before

After:
after


erikquinn:
@Fechulo What is that "reliable source" for navdata? Just curious

Fechulo:
@erikquinn I found [these](http://nav.vatsim-germany.org/) files, used for the ATC clients for VATSIM. It isn't exactly reliable as it isn't real life data, but the information is extremely accurate.

erikquinn:
Oh okay, cool. I'm always looking for more data sources. Usually I get everything from SkyVector, they seem to be pretty spot on, but even they don't have everything...

Fechulo:
It's not easy to find...

PR #490, opened 2016-05-12
Fix wiii wimm star

ghariman:
Minor bug for WIII star entry points and headings.

Also followed Erik's advice on how to use Github branches better to avoid showing old historically commits.


erikquinn:
@ghariman Thank you! :smiley:

PR #492, opened 2016-05-13
Add Polygonal Boundaries at EDDT

Fechulo:
This PR also includes some tweaking to the arrival traffic.


erikquinn:
@Fechulo In replacing all `wave` arrivals with `surge` arrivals (to stay consistent with previous traffic flows, I essentially undid some of the changes to the arrival traffic you made here via #493. Sorry, and you may want to play with it briefly, and adjust if necessary. It should hopefully be a pretty close translation to what it was originally.

Fechulo:
I was, recently playing at EDDM, where the traffic is very manageable and not stressing at all. At that point I realized that in most of the airports I've made, the traffic is very difficult to control and and it certainly stresses me a bit. So I decided that for future airports I would try to lower down the amount of waves and amount of traffic in general. That's why I did those changes to EDDT. Don't worry about, I actually wanted to play around with the new flows, to see if I wanted to change something.

Issue #483, opened 2016-05-10
Seems to crash when playing with speed multiplier for extended periods


anishthecoder:
I've tried playing this in multiple browsers, and any time I use the speed multipliers for too long, the game eventually hangs. I don't see any console errors, but the entire interface becomes unresponsive...

However, really really great game!

Fechulo:
Does this happen in all airports or in a specific one? If so, which airport?

anishthecoder:
I've tried this on a few different of airports and it seemed to happen in each of the cases: San Francisco (primarily), LAX, and Heathrow. If I ever used the speed multiplier (even just a couple of times), it would hang before I could clear around 20 flights. I'm trying it out without the speed multiplier and will let you know if it still crashes.

Fechulo:
thanks, I'll look into it!

erikquinn:
@anishthecoder By chance, does this happen as an aircraft is in the process of trying to join the localizer? (I have a feeling it may be `aircraft.js:2217`, as `this.rwy_dep` will be null for arrivals)

anishthecoder:
@erikquinn Was trying to verify that, but having trouble because of a new issue #488

anishthecoder:
@erikquinn Yes, this does appear to happen in the process of trying to join the localizer (but it could just be my perception, there's a lot happening in traffic when it does crash)

erikquinn:
@anishthecoder Okay, so maybe that is what's causing it. And now that I think about it, that could make sense. Perhaps using the speed multiplier is causing it to be too far away to start the turn at one time, and at the next update, be too far past the point where they should have turned, thus exacerbating the problem. _This could be it_.

I made the puny little change I described above. At your leisure, please try out this link and advise if the problem is fixed there (while I sit here and cross my fingers, hoping it is this easy :wink:)

[http://erikquinn.github.io/atc/b/483-test](http://erikquinn.github.io/atc/b/483-test)

anishthecoder:
@erikquinn Sadly not.... :smile: Just tried and it hung again. This time though, it seemed to happen just when one of the landing crafts was ready to switch to ground (or it looked like that spot on the runway where it's ready to switch to ground). There were no new localizer capture attempts.

anishthecoder:
@erikquinn Surprisingly, when I left the window hung for a while and zoomed in and out using my scroll wheel, things started working again.... :open_mouth: and then hung again when I tried the speed multiplier again... this time with no recovery

[...]

Issue #477, opened 2016-05-06
IE & Safari Not Working


erikquinn:
Imagine my surprise when I noticed that "everybody's favorite browser" is having troubles again... It seems to not be loading airports somehow, which leads to it failing to start the game. :angry: 

It dies on `ui.js:113` when it can't find `prop.airport.airports`, because `prop` does not have a property `airport`. Do the same thing in Chrome or Firefox, and all the airports are there.

![image](https://cloud.githubusercontent.com/assets/5103735/15080487/16e014a2-1384-11e6-9531-74d403312828.png)

![image](https://cloud.githubusercontent.com/assets/5103735/15080537/46915670-1384-11e6-997f-1ccd06cd797a.png)

tedrek:
It would be very useful to specify what version of IE this is affecting.  (I have limited access to IE and rarely touch it)

avdg:
This error can't yet be reduced to a smaller use case. Can't it? Maybe adding conditional console.log (Error.stack) or debugger; statement can help?

avdg:
This bug doesn't occur in edge.

avdg:
In ie 11, this line of code is causing an error:

https://github.com/zlsa/atc/blob/e985d0bc8d2ca678366805cded9038b61c691430/assets/scripts/airport.js#L735

I'm not sure if that is causing a snowball effect or not (making the value undefined), and why ie 11 didn't stop executing.

erikquinn:
This is also the case in Safari.

PR #493, opened 2016-05-14
STARs and Minor Misc Fixes

### LINKED ISSUE #485: Standard Terminal Arrival Routes
We've got SIDs, and #464 reworked some things to improve upon the fms, and the next logical step is to add STARs with the same format as the SIDs. Should be pretty quick and easy work.

erikquinn:
Resolves #485.

Enables STARs under the same basic format as SIDs, and fixes a few other items I came across, such as:
- Localizer draw distance not being drawn separately for each end
- Undoes my previous removal of the circular reference between `Aircraft` and `FlightManagmentSystem`, who both have each other as a property (apparently this is not an issue in modern JS, and so I put it back to the way it was, as it is much easier to work with). Verified this does _not_ adversely affect performance.
- Adds various improvements to arrival spawning methods (now three methods available: `cyclic`, `wave` (different than it used to be!), and `surge` (new version of the old "wave", that takes more useful parameters: `frequency`, and `entrail`, where you specify the entrail distance during the heavy part of the push, and the math does the rest to achieve the desired frequency)
- Adds explanations and diagrams of arrival spawning methods
- Adds logic that checks to make sure the fixes referenced in an airport are all actually present in the fix listing (logs warnings)

Also worth noting that aircraft will now spawn at the first point along the STAR if spawning via stars. See `ksfo.json` to see how entries within the `arrivals` section is changed to acommodate spawning along a STAR.


jakobeng1303:
I suppose the planes won't descend without altitude restrictions, will they?

Alpi-no:
The demo is not working in safari!
It stops loading the airport when the plane rj85 is being loaded. I don't know why. It runs on Chrome tough. 
When trying to start, I figured out, that taxiing wouldn't work. Not taxi, nor wait. I wasn't able to send "taxi" to the plane. When typing taxi runway 28R/L//(0)1R/L the game would put out the message "callsign, no runway RUNWAY xx". At least it did react to the message and I was able to "send" it. And it's pretty laggy! And 5 mins after the start of the game, it crashes! (I played it on max speed though)

erikquinn:
In investigating this^, I did stumble across one thing where issuing `caf` **before** taxiing will cause a false positive readback:
image

---

@jakobeng1303 The Safari issue is actually unrelated to this, as you'll see if you try to go to the normal [zlsa.github.io/atc](http://zlsa.github.io/atc) in Safari. That is related to a known issue, #477, which I updated to include Safari as well.

Taxiing is working fine from what I'm seeing...

> When typing taxi runway 28R

Perhaps this is the issue; are you actually typing `taxi runway 28R`? If so, that would do it, as the correct command would be simply `taxi 28r`.

Ref. the last problem about timewarping, that is also pre-existing and being investigated in issue #483. If you could describe in greater detail with screenshots and/or console messages and/or s screen recording how that is happening, please head over to #483 and put your comments there so we can get it cleaned up.

But as it stands, I don't believe any of the above are related to this PR in particular.

Alpi-no:
@erikquinn You are absolutely right with the taxiing. My bad! 
The problem with the other things is that all of them only occur in your demo, not in the regular game. That's why I thought, it might be helpful!

Fechulo:
This is great! I would like to know exactly how surge and wave work, but apart from that, I say merge it! If/When we decide to have a center control mode, a 'star' command might be useful to have, however we don't need it right now.

Alpi-no:
@Fechulo I think that a star command might be very useful right now because it suits the purpose better than just `route` or `rr`. Another way could be `callsign c(leared) xx (star)`

Fechulo:
@jakobeng1303  I said that it wasn't useful right now, because aircraft enter the airspace already following a star. In the case of Vienna you might be confusing stars with RNAV transitions. regarding the alternative command, I'd  personally stick with [callsign] STAR [name of STAR], since the command [callsign] C [name of STAR] might have conflicts with the 'CLIMB' / 'C' command.

Alpi-no:
@Fechulo that is probably the best one

erikquinn:
Okay, if it seems good to you guys, we can go ahead. :smile: And then I think it'll be time to update the documentation to reflect the changes and tackle the pile of small bugs that have come up lately... (yawn!)

@Fechulo there are some thorough descriptions in the code, see `airport.js` in each of the `zlsa.atc.Arrivalxxxxxxx` definitions. I'll work on getting the other documentation in sync with what's shown there, but there should be enough detail there to outline it. Basically, the old "wave" is now called "surge" (but takes different parameters), and the new "wave" is just like "cyclic", but instead of rate of change of the arrival rate being _constant_, it will oscillate from the highest rate to the lowest rate in a sinusoidal pattern. There are pictures that explain it better in commented-out code.

Issue #230, opened 2015-11-10
Active Runway


Maverick283:
In real life, the ATC determines the active runway(s) based upon the winds current at the airport.

When introducing variable winds, I'd suggest an input where the user can specify the active runway. 

Once the SIDs and so on got going and are all figured out, I believe it'd be an advantage to bind SIDs to runways. Then, whenever the active runway is chosen, only the necessary information could be drawn on the canvas, like the SIDs for the runway.

Also the aircraft could taxi to the active runways when selected by the user, and the user will get in trouble when he forgets officially changing it as the departing traffic would stand on the wrong side.

zlsa:
At the moment, you can instruct an aircraft to `taxi 21L` if so desired.

Maverick283:
Correct, but when not specifying where the aircraft is supposed to taxi to, it will taxi to some runway. This behaviour could be changed so it would taxi to the active runway. This would save a lot of typing.

zlsa:
The active runway is already determined internally by the wind IIRC.

Maverick283:
Yeah.... Usually the ATCs decision ;)

erikquinn:
Covered by #658. Closing.

PR #494, opened 2016-05-15
New Airport: KSFO South East Ops

ghariman:
Created new airport called "KSFOSE" for KSFO in South East Operations (takeoffs 10L/R, landings 19L/R).

Tested on ghariman.github.io/atctest/

southeast-plan


erikquinn:
Hi @ghariman-- unless there's a consensus to the contrary, I don't think adding new "airports" in order to utilize a different runway configuration is the best path to achieving this. I want the same thing you do (to be able to change it up with different flow configurations), but I'm not inclined to merge this myself, as I would rather see us get the game to change the wind dynamically, or at least allow the user to "set" the wind, thus determining the runway configurations.

I've actually been wanting changing/changeable wind for quite a while now, perhaps work on dynamic wind patterns should be moved forward to fulfill the intent of this pull... related issues where wind was discussed include #68, #230, and PR #169

Alpi-no:
@erikquinn I hope this questions doesn't sound too stupid, but is it that hard to implement a box popping up when choosing an airport that asks for windspeed and direction, maybe with a notice about the runway use with certain winds.

erikquinn:
Probably not, but I straight-up suck at CSS and UI manipulation. I only really have experience and value in the more back-end parts. We do need to figure out how that would all work though, because if an en route mode is added, then there's different "facility" types (En Route, TRACON, and eventually Tower), and then each of those would have a list... when do you show each list? How do you present the choices to the user? What is the wind like? What type of theme should your scope be using, so it looks like something you're used to? There just become a lot of questions, and I'm not sure the best way to set that up from a UX perspective. It's above my paygrade. :stuck_out_tongue_winking_eye: But I'm sure there's a good way to give us the choices with regard to wind, it's just not a task for me personally.

Alpi-no:
@erikquinn sadly my programming skills are not existent 😆 so I can input only some ideas for this topic

erikquinn:
This can be revisited in the future when we deal with changing runway configurations during a session. Closing.

PR #495, opened 2016-05-15
Fixes #475 (Restricted Areas Not Drawn)

### LINKED ISSUE #475: Restricted areas not loaded correctly
There seems to be a bug, where only the first restricted area of an airport is loaded and the rest are skipped. No error is shown in the log and I suppose it may have something to do with the polygonal airspace boundaries.

Fechulo:
the cc.clip() was preventing the canvas from drawing all restricted areas but the first. Deleting this line doesn't seem to affect the range rings. This PR fixes #475


Fechulo:
I just realized that, the cc.clip() was actually important for the airports with polygonal airspace.

Fechulo:
This works now, restricted areas are drawn and range rings don't go outside the airspace!

erikquinn:
looks good... good find :+1:

PR #496, opened 2016-05-16
Added some fixes at LOWW

maebruck:
Added some fixes at LOWW


maebruck:
Reopened pull request although some fixes are outside of loww airspace, they are needed for SIDs und STARs

erikquinn:
@lachsschinken124 Unless we are actually using these fixes for something, to be hoenst, this only seems to add clutter at an airport that already has an extremely high number of depicted fixes... I believe @Fechulo originally had all of these fixes in there, and removed the ones that weren't needed for this exact reason.

If you want to add the SIDs/STARs that use these fixes, it may be worth adding... But for now, I'm afraid I don't personally see how adding _just the fixes_ by themselves is really an improvement.

### Current:

before

### Proposed:

after

Fechulo:
@erikquinn as far as I know, I've added all the SID's, so I doubt those fixes are for SID's. I would personally keep the new fixes for STAR's (eg, WW191 and WW192), but I don't really see the point of having more unnecessary fixes.

erikquinn:
@Fechulo Okay, I see. And yeah, if the fixes come with STARs, I have no problems merging them. But each fix that's added needs to have a purpose, as we've already got quite a few.

Fechulo:
> quite a few.

lol. In fact, There's many more fixes for the STAR's that could be added, which wouldn't clutter the airspace.

maebruck:
I'll close this pull request for now

PR #497, opened 2016-05-16
More fixes and Stars for LOWW

Alpi-no:
I don't expect it to be merged right now, but as I think of it being pretty much done, I just wanted to put it here!


erikquinn:
Holy cow-- at first glance, these STARs look **super** complicated... thanks for putting in the time :)

This will obviously depend on #493, which is in need of some community feedback. Once it goes, then we can have a look at this! :+1:

assets/airports/loww.json
```diff
@@ -335,6 +407,59 @@
...
+        "16": [["PESAT", "A60+|S250"], ["WW684", "A60+|S250"], ["WW682", "A60+|S250"], ["WW680", "A60+|S250"], ["WW678", "A50+|S250"], ["WW676", "A50+|S250"], ["WW674", "A50+|S250"], ["WW673", "A50+|S250"], ["WW672", "A50+|S250"], ["WW671", "A50+|S250"], ["WW670", "A50|S250"]],
+        "34": [["PESAT", "A60+|S250"], ["WW991", "A60+|S250"], ["WW989", "A60+|S250"], ["WW981", "A60+|S250"], ["WW979", "A60+|S250"], ["WW977", "A60+|S250"], ["WW975", "A60+|S250"], ["WW974", "A60+|S250"], ["WW973", "A40+|S250"], ["WW972", "A40+|S250"], ["WW971", "A30+|S250"], ["WW970", "A30|S250"]]
+      },
+      "body": [],
+      "draw": [["PESAT", "WW667", "WW665", "WW663", "WW661", "WW659", "WW657", "WW655", "WW653", "WW652", "WW651", "WW650"], ["PESAT", "WW966", "WW956", "WW954", "WW953", "WW952", "WW951", "WW950"], ["PESAT", "WW684", "WW682", "WW680", "WW678", "WW676", "WW674", "WW673", "WW672", "WW671", "WW670"], ["PESAT", "WW991", "WW989", "WW981", "WW979", "WW977", "WW975", "WW974", "WW973", "WW972", "WW971", "WW970"]]
+    }
```
erikquinn, line 462: one thing jumps out here-- missing a closing bracket and comma at the end here, which would cause the game to fail the load of `loww.json`.

Alpi-no:
@erikquinn I hope I didn't make any mistakes, but they should be working fine 😆 I think rather than you thanking me, it should be me and the community thanking you for all the effort you put into this project!!! Great game!

erikquinn:
And the `arrivals` section will need to be updated to get them to actually fly the STARs when they spawn. Feel free to reference [ksfo.json:241](https://github.com/zlsa/atc/pull/493/files#diff-03dba5225c71b7b1f30dcfbd94955dd6R308), the example airport included in #493 when adding the stars to the spawn patterns!

erikquinn:
@jakobeng1303 ohhhh this project is much bigger than any single person, there are quite a few people to have done more than me. In fact, I'm kind of late to the party... :disappointed: But, now that I've found this awesome game, I'm here to stay and help as much as possible. :smile:

assets/airports/loww.json
Alpi-no, line 462: @erikquinn you are absolutely right. Totally missed that!

Alpi-no:
@erikquinn I don't know for how long I'll stay, but I'll give it a try and help this incredible project! 
Will `"route": "xx",` clear the plane right form the beginning to a certain star or what will that command do?

erikquinn:
@jakobeng1303 it describes the route of flight. So the beginning of the star, the star id, the destination airport, eg `route: "MXW.PYE1.KSFO"` will cause the aircraft to spawn at MXW, follow and descend via the PYE1 STAR, with a filed destination of KSFO. (Reasoning for including the last part is because we should eventually be adding capability to work arrivals/departures to/from multiple airports within a single airspace)

Alpi-no:
@erikquinn so in my case adding in `route: Mabod` isn't enough, but I'd need MABOD.MABOD4K.LOWW ? basically I have to add in those for all 4 possible runways. How do planes know, which one to take?

Fechulo:
@jakobeng1303  As I said in #493 I think you might be confusing STAR's with RNAV transitions.

Here are the STAR's (e.g., MIKOV 7W) which end in the point MABOD:
trans

And here is the transition (MABOD4K) which takes aircraft form MABOD to the ILS (in this case, RWY11):
star

I'm not exactly sure what the route would look like, but it would be something like this: MIKOV.MIKOV7W.MABOD.MABOD4K.LOWW, so the STAR shouldn't start at MABOD, it should start at MIKOV

Alpi-no:
@Fechulo Damn. You are absolutely right! I'm confusing this and this makes me feel very stupid 😆. Just to confirm, is it planned to just add the star, or the transitions as well? 
The Problem with Mikov7W is, that currently Mabod is the only waypoint we have, so planes would spawn into Mabod, not Mikov.

Fechulo:
@jakobeng1303  don't feel stupid, a lot of people make this mistake! Currently, aircraft do follow the MIKOV7W, as you can see in the picture:
mikov

I'm not sure how the transitions could be assigned, but maybe, we should assume that MIKOV 7W is actually part of the flight plan and MABOD 4K is the STAR. It's not very realistic, but it's probably good enough.

Alpi-no:
@Fechulo I think for LOWW (currently) clearing the transition will be more helpful but in the future we/I might add it. If I got you right, that would mean changing all the arrivals to their actual beginning eg Mikov? how would I clear a plane then? Mikov7w and then mabod4k or at the same time?

Fechulo:
Right now, it's only necessary to clear them for MABOD 4K, since the are already following MIKOV 7W when they enter the airspace. However, if/when we have en route control, we might need to clear the for both, but we'll leave that to future contributors.

Alpi-no:
@Fechulo thanks for all the information, something I'm still trying to figure out, is how I'm supposed to write that route:  `MABOD.MABOD.LOWW` or else? I have left out the 4k on purpose because from my pov it didn't seem logic to tell a plane to use the Mabod4k transition, when it's actually flying the 5k. I hope it's clear what I'm trying to say

Fechulo:
@jakobeng1303 I'm afraid it isn't clear what you're talking about :smile: , at least I have no idea what you are referring to, sorry...

erikquinn:
@jakobeng1303 Actually, now that I looked a bit more closely at the diff of this pull, there is a critical difference in nomenclature in the USA vs Europe...

The STARs are implemented in the FAA convention of `transition.procedure.destination`, where transition is _not_ the same transition you guys are talking about. The word "transition" to me means the point that links a terminal procedure (SID/STAR) with the en route structure (airways). So you fly on an airway to one of a few "transitions" that connect you to the STAR, and then that takes you to the airport.

In Europe, it seems to me like it goes like this: you follow an airway to some fix that connects you to the STAR. (In the USA, I would call that fix the "transition", and Europe doesn't really name it). You then follow the STAR to the fix for which the STAR is named, at which point, you are still a ways away from the airport, and must be assigned a "transition" that connects you from the end of the STAR to a particular runway. Thus, the "transition" is actually kind of like a STAR that comes after the STAR.

I now see in this PR, you listed the _runways_ within `transitions`, and that will not work. For it to work, the transitions need to be the fixes from which arrivals can join the STAR (even if only one). If multiple (european) "transitions" off of this STAR are needed, multiple STARs will need to be created. The best course of action (at least for now) is probably to pick only one (european) "transition" to whichever runway is in use based on the wind, and we can figure the rest out later.

Alpi-no:
@erikquinn I should have looked that up earlier :/ that is quite a big misunderstanding. Luckily we know now!

Alpi-no:
@erikquinn so the European format would be like:
`"stars": {
    "NERDU 4K": {
      "icao": "NERDU4K",
      "name": "Nerdu4K"
      "body": [["NERDU", "A60+|S250"], ["WW666", "A50+|S250"], ["WW664", "A50+|S250"], ["WW658",       "A50+|S250"], ["WW656", "A50+|S250"], ["WW654", "A50+|S250"], ["WW653", "A50+|S250"], ["WW652",    "A50+|S250"], ["WW651", "A50+|S250"], ["WW650", "A50|S250"]],
      "draw": ["NERDU", "WW666", "WW664", "WW658", "WW656", "WW654", "WW653", "WW652",      "WW651", "WW650"],`

Is that kind of correct?

Alpi-no:
for some reason it clutters all together -.- 
"stars": {
    "NERDU 4K": {
      "icao": "NERDU4K",
      "name": "Nerdu4K"
      "body": [["NERDU", "A60+|S250"], ["WW666", "A50+|S250"], ["WW664", "A50+|S250"], ["WW658",       "A50+|S250"], ["WW656", "A50+|S250"], ["WW654", "A50+|S250"], ["WW653", "A50+|S250"], ["WW652",    "A50+|S250"], ["WW651", "A50+|S250"], ["WW650", "A50|S250"]],
      "draw": ["NERDU", "WW666", "WW664", "WW658", "WW656", "WW654", "WW653", "WW652",      "WW651", "WW650"],

erikquinn:
Try using triple-backticks above and below your code, and it should indent properly.

```
"NERDU4K": {
  "icao": "NERDU4K",
  "name": "Nerdu 4K",
  "body": [["NERDU", "A60+|S250"], ["WW666", "A50+|S250"], ["WW664", "A50+|S250"], ["WW658", "A50+|S250"], ["WW656", "A50+|S250"], ["WW654", "A50+|S250"], ["WW653", "A50+|S250"], ["WW652", "A50+|S250"], ["WW651", "A50+|S250"], ["WW650", "A50|S250"]],
  "draw": ["NERDU", "WW666", "WW664", "WW658", "WW656", "WW654", "WW653", "WW652", "WW651", "WW650"]
}
```

The property name (eg what you put as the label for the whole block) needs to match the ICAO identifier. The `name` property is just to help the speech synthesis know how to pronounce it since she's not quite smart enough to get "Brookmans Park" out of just "BPK". :wink: Rest looks fine syntactically; as long as it matches the chart it should be correct.

erikquinn:
@jakobeng1303 Oh, and also remove the comma at the end of the `draw` section, since it is the last element. And add one to the end of `name`, since it is not.

Alpi-no:
@erikquinn thanks for that & makes sense

Alpi-no:
I've decided to close this pull for now and make a new one as soon as I have converted all the transitions into stars or whatever 😆 I hope that'll work though!

PR #499, opened 2016-05-16
New Airport: EGKK (London Gatwick)

Fechulo:
Gatwick Airport is the second busiest airport in the UK, after Heathrow. However, we must give it recoginition, as it operates with only one runway (RWY26R/08L) You might notice that it's 26R and not just 26. This is because, there are actually two runways, but 26L/08R is too short and to close to the other runway, so it is now used as a taxiway. This airport includes, restricted areas, polygonal airspace, two new airlines: Thomas Cook Airlines and Thomson Airways and most importantly, I took the time to organize the fixes in alphabetical order. I hope you like it!


erikquinn:
@Fechulo The OCD in me thanks you for that! :laughing:

erikquinn:
werks perrty good! :smile:

PR #505, opened 2016-05-17
Remove arrow functions and spread operator for IE compatibility

### LINKED ISSUE #477: IE & Safari Not Working

erikquinn:
Resolves #477, _at least_ the Internet Explorer portion. Was due to the use of one arrow function and one use of the spread operator, which is not yet cool with IE.

I _assume_ Safari's problem is the same, but I am unable to verify it is 100% fixed. What I can say is that the issue described here no longer happens. In Safari, the game now loads completely, but doesn't display the airport (blank green) for the version I have. This may not be a fair test though, because I am on windows, and only can get Safari 5.1.7 (about four years old, v9.1 is current). Since it loads, I imagine the problem is the same as it was with IE, and _should_ theoretically be working now.

### DIFF
--- a/assets/scripts/airport.js
+++ b/assets/scripts/airport.js
@@ -693,7 +693,7 @@ var Airport=Fiber.extend(function() {
         var pos = new Position(this.perimeter.poly[0].position, this.position, this.magnetic_north);
         var len = nm(vlen(vsub(pos.position, this.position.position)));
         var apt = this;
-        this.ctr_radius = Math.max(...$.map(this.perimeter.poly, function(v) {return vlen(vsub(v.position,new Position(apt.rr_center, apt.position, apt.magnetic_north).position));}));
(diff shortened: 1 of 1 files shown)


Alpi-no:
@erikquinn I just tested Safari and it still can't load the aircraft rj85 :/

erikquinn:
@jakobeng1303 Theres no change until this actually gets merged, so that's expected. I can push the branch to my gh-pages to create a demo link though. I'll do that in a bit.

erikquinn:
Sorry for the delay. 

Demo link: [erikquinn.github.io/atc/b/ie-safari](http://erikquinn.github.io/atc/b/ie-safari)

@jakobeng1303 You have Safari, would you mind giving that one a try? All it has to do it load the game, and we should be good.

Alpi-no:
@erikquinn It works just fine!

erikquinn:
good news. thank you :smile:

PR #517, opened 2016-05-19
Add Polygonal Airspace at ENGM

Fechulo:
-Added the Oslo TMA
-Removed restricted area D106 (it's outside the Oslo TMA)


erikquinn:
Can't quite find it, but something in here is breaking ENGM.

Fechulo:
mmh, I'll check

Fechulo:
It's working for me, is it just not loading?

erikquinn:
Nah, it works. It just dies when it tries to run `vsub()`, which indicates to me that it's probably something in the airspace boundary.

image

erikquinn:
If you timewarp by like 100 or so, you get it in a few seconds.

Fechulo:
I was testing it on an old version, let me try again.

Fechulo:
Found it! 
I accidentally removed fix "_ENGM103060", which was being used for the ESEBA STAR.
image

I'm not sure what this fix is for, which is probably the reason why I deleted it.

erikquinn:
All better :) Thanks!

PR #524, opened 2016-05-21
[completed] VIDP, Indira Gandhi International (Delhi)

eliuuk:
Indira Gandhi International serves as the largest airport in India, and 25th in the world in 2015. It is the airport for the capital city of India, Delhi.
- [x] Add all waypoints required for SIDS and STARS (295 waypoints!)
- [x] Script the SIDS
- [x] Script the STARS
- [x] Attempt a polygonal airspace
- [x] Create Air India
- [x] Create Air India Express
- [x] Create Air India Regional
- [x] Create AirAsia India
- [x] Create AirAsia X
- [x] Create SpiceJet
- [x] Create IndiGo
- [x] Create Jet Airways
- [x] Create GoAir
- [x] Create Vistara
- [x] Create the AT72-600
- [x] Edit the load file for the airlines
- [x] Add spawns for the new airlines
- [x] Add ATR76 to the load file


erikquinn:
Looking good!

I have two comments:
1. Would you mind removing (or at least hiding) the RNAV fixes that aren't in use? There's quite a lot of them, and if they aren't needed, it's probably better not to display them. You can just delete the unneeded ones, or make them "hidden fixes" by prepending an underscore to their names (eg `"D340N" --> "_D340N"`), and they will be usable, but not drawn on the scope.
   image
2. In reference to issue #526, it's likely that the airspace boundary should be extended from what you have included here. If you're unable to find the dimensions of the outer area, then perhaps we should just make the airspace a circle here? Just something to avoid aircraft leaving/re-entering the airspace and invoking multiple callups.

eliuuk:
@erikquinn Thanks, I've removed some more waypoints - not sure if that's all of them, since there is a point where the unused and used waypoints split.

Concerning the airspace, I've edited it, might be a bit fictional but solves the problem!

PR #530, opened 2016-05-25
Add missing fixes to some airports - Fixes #529

### LINKED ISSUE #529: I accidentally removed necessary fixes
As the title suggests, I may have deleted some fixes which were used by arrivals in many airports, which is causing the game to crash. I messed up, so I'll fix it :smile:

Fechulo:
I'm not sure what these missing fixes are for, but I added them back, so that the game doesn't crash. This PR fixes #529


erikquinn:
I added them when I went through all the airports following changes to the way aircraft are spawned. Since they now spawn at the first fix in the list, [for example](https://github.com/Fechulo/atc/blob/a7656268ebfe387f1359b3aaf36048a9bec6f918/assets/airports/eidw.json#L1049):

```
    {
      "_name":  "OSGAR 1L",
      "speed":    250,
      "fixes":   ["_EIDW243060", "OSGAR", "DIRUM", "KEPOR", "ARVOK", "SORIN", "SIVNA", "SUGAD", "DW704", "DW705", "DW706", "SOPEP", "LAPMO"],
      "airlines":  [
        ["ein", 4],
        ["aal", 3],
        ["aca", 2],
        ["dal", 2],
        ["ual", 2]
      ],
      "type": "random",
      "offset": 20,
      "frequency": 5,
      "altitude":  10000
    },
```

In that example, if I left it as it was, the arrivals would spawn right AT `OSGAR`, which is _RIGHT_ on the border of the airspace...
image

Instead, I created these fix-radial-distance points that would cause them to spawn sufficiently outside of the airspace, but not super far away either. Basically, it was my lazy alternative to looking up all the charts for all the airports.

PR #532, opened 2016-05-26
New Airport: TNCM (Sint Maarten Princess Juliana)

Fechulo:
Princess Juliana International Airport is probably the most famous airport in the world, due to its amazing approach over Maho beach. I wanted to add the new STAR format, however it didn't really makes sense, since the only actual STAR in TNCM is to fly a heading after the initial fix:
vectors
Instead, I decided to add the RNAV (GNSS) approach for RWY10 as the STAR. Because it's not a STAR, it doesn't have a name or identifier, so I used the old format:
gnss
I also wanted to add the terrain around Sint Maarten, as this is the reason why aircraft have to turn right immediately after take off. The problem is I don't know how to do it :smile: . @culebron I would love it if you gave us a step-by-step guide on how to add terrain!


erikquinn:
@culebron I've tried my hand at terrain a few times, and was able to make some headway, but I don't see how editing nodes can fix what I've got pictured below. If you could briefly provide some guidance on that, I think I can figure out the rest, and will gladly help with the write-up on how to do the rest of it (I believe all the difficult work is behind me, and it's just tedious cleaning from here on out). Do you have any advice on how to clean this up?

image

erikquinn:
Looks sweet, no problems encountered. :smile: And I created a new issue for the terrain stuff so the above is not lost to the abyss of the internet.

PR #537, opened 2016-05-28
Show aircraft's full route on hover of strip

erikquinn:
What about something like this to address #407? Hovering over the strip will show the route?

image
Note: Don't mind the white scribble, the screen capture always excludes the mouse cursor, so I drew it in to show where the cursor was.

Thoughts, anyone? I'm not sure if I really like it. More ideal might be a right-click that toggles the strip expanding a bit in height to show the route on a third, smaller row... I just don't know if I can achieve the latter...


Fechulo:
Even though I like it, I prefer the idea of the strip expanding to show the route. I also think this would limit the usage of the 'Say Route' command, which is already quite useless. I'm not sure if the 'sr' command was originally meant as a temporary solution, but this feature would be completely replacing it.

eliuuk:
I prefer this in comparison to the 'sr' command, it's just a whole lot easier. Plus, it saves the woman on the voiceover from speaking too much - she already talks extremely slow.

Fechulo:
Haha, poor woman! I agree, it's a better option, I just wanted to now if @erikquinn 's intentions were to eventually get rid of the 'sr' command.

erikquinn:
Oh yeah, it was 100% a temporary solution. It can get the axe. lol

Thanks for the feedback guys. I think I agree, just wanted to toss this up here and get your thoughts. I'll close this now, and we can find another solution sometime.

EQ

PR #540, opened 2016-05-28
Removed TAP Portugal from EGCC

### LINKED ISSUE #539: EGCC:: TAP does not exist
TAP Portugal doesn't exist, and for some reason I added it in the arrivals.
#540

eliuuk:
Fixes the crash when the game cannot find TAP. Fixes #539


erikquinn:
@indianbhaji I'm actually not seeing any changes here... it appears that all this PR would do is remove whitespace. It doesn't actually remove TAP from `egcc.json:404`.

eliuuk:
Weird... I'll repush this too.

PR #541, opened 2016-05-31
Update LOWW

Alpi-no:
demo: http://lachsschinken124.github.io/atc/ ATTENTION: in the demo stars will only work as a routing function, altitude or speed restrictions are (I don't know why) not working. 
charts for the transitions (in game stars) can be obtained here: http://www.vacc-austria.org/index.php?page=content/chartlist&icao=LOWW
There are restricted zones for the LHCC and LZBB FIR. Those are in place because these areas are outside of Austria, but still have to be controlled by LOWW APP.


erikquinn:
@jakobeng1303 Is it your intent that the arrivals are not assigned the STARs, and that they just fly the series of fixes? Just making sure.

The altitude and speed restrictions are working for me. Perhaps make sure you actually tell them to descend via the star, as that is a separate action from assigning it. Use the `dvs` command after the star has been assigned, and they should slow/descend to the assigned speed/altitude. Note that if an "at/above" restrcition is given for the next fix, the aircraft won't descend because it isn't _required_ to comply with the sid. Only when there is an "AT" or "AT/BELOW" restriction will you see the aircraft begin the descent. It's not the greatest setup, but it's what we've got for now.

I like the restricted areas. Overall, it's looking and working quite well now! Just let me know about the thoughts I had above.

Alpi-no:
@erikquinn The planes are being assigned the star, it's just that I do it. Like in real life I clear their transition. It wouldn't make sense to save the star in the `arrival` section, because which star is going to be used, is individual. eg one might favor arr 11 and dep 16 but another one likes arr 34 and dep 29. Assigned a plane to star immediately after spawning wouldn't allow that. 

Thanks for telling me that! I didn't know that command existed!

PR #544, opened 2016-06-01
Fixes #542

### LINKED ISSUE #542: Fixes do not re-appear when changing airports
If you hide fixes with the 'toggle fixes' button, in airports like KSEA, the fixes and runway labels will disappear, as expected. However, if you then switch to another airport, the fixes remain hidden. Ideally, we want fixes to be drawn whenever an airport is loaded, or always draw the fixes in airports that don't have video maps. I think I'm able to fix this myself, but I'm not at home, so I can't test any changes.

Fechulo:
This PR fixes the issue described in #542 . I didn't explain it very clearly, so I'm gonna try to explain it here. The airports that have video maps, look very hideous because of all the fixes and labels being drawn: 
part 1
Thankfully, we can toggle the drawing of fixes and labels, which looks MUCH better:
part 2
The problem comes when the toggle fixes and labels button is disabled and we then switch to an airport which doesn't contain a video map. The issue is that fixes and labels are not drawn and there's no button to change that:
part 3
This PR simply adds a line, which draws fixes and labels by default every time you load an airport

### DIFF
--- a/assets/scripts/airport.js
+++ b/assets/scripts/airport.js
@@ -1232,6 +1232,7 @@ function airport_set(icao) {
     .text(prop.airport.current.icao.toUpperCase())
     .attr("title", airport.name);
 
+  prop.canvas.draw_labels = true;
   $('.toggle-labels').toggle(
     !$.isEmptyObject(prop.airport.current.maps));
 


erikquinn:
.......oops. :laughing: 

thanks @Fechulo!

PR #555, opened 2016-06-04
New Airport: KBOS (Boston Logan Airport)

Fechulo:
This is my first take on an American airport; General Edward Lawrence Logan International Airport or simply Boston Logan Airport. Of course, all fixes are ordered alphabetically! :smile: @erikquinn Can you confirm that the airspace is correct? And if it isn't too much to ask, would you be able to add the video map for the area?
Thanks in advance!


erikquinn:
It looks about right from what I vaguely remember, but I don't have the files for it, and the vZBW site is down, so I can't download them. I can ask some people and try to get a hold of the files we need. For now, it seems close enough. They've got some busy looking SIDs up at Boston, don't they??? lol

Fechulo:
Thanks, don't worry about the video map. They are almost as busy as LOWW!

PR #559, opened 2016-06-06
PCT: RVM & Airspace (Mt Vernon Area)

erikquinn:
Adds polygonal boundary (I guessed at it... don't have maps that would show me for sure) and radar video map for the Mount Vernon area of Potomac TRACON (the area where KDCA lies). Was quite annoying, because the source data was _VERY_ highly detailed, and had to do some more qGIS learning to be able to reduce it sufficiently (from 37k lines to ~6k). Looks good though! Only thing is that seeing Dulles hanging off the west side makes me want that airport too... :stuck_out_tongue_winking_eye: 

image


Fechulo:
I'd be willing to add Dulles, but you're gonna have to wait until I finish my exams...

erikquinn:
@Fechulo lol really what needs to happen is for all this freezing business to get resolved and everything to stabilize again so I can go after some new functionality, like servicing multiple airports in the same simulation. It just isn't wise to do it until she's nice and stable. :cry:

Issue #402, opened 2016-02-20
Merge aircraft jsons to single file per manufacturer


MaicomMR:
Hello budyes.
I'm think the possibility of merge all aircrafts for a single .json file, I believe that will give a lot of work, but will reduce a significant time of "loading", with this change we can create a lot of more detailed aircrafts to the game without increase the load time.
Unfortunately I do not know much about java programming to try Crair something, I leave here only an idea (I believe that many already have), but here is official and open for discussion.

erikquinn:
Unless I'm mistaken, I dont believe merging json files will improve the performance... it still has to load all the same data, it just doesn't have to jump around between different files, which should be <1% of the clock cycles being used in the load. That said, I have noticed an abnormally long wait for each aircraft to load, which may be a problem with gihub pages servers being slow. Not sure.

I absolutely agree, however, that we can do more to reduce the loading time... I was thinking we ought to do so by _only loading the files we need_. Currently, we just load everything we have. Instead, perhaps we should change the `aircraft_init()` function to be more intelligent, checking the selected airport to see what to load. Then the initial load of the game would be much faster, as it would only load the current airport, any terrain for that airport, and only the aircraft that will arrive or depart from that airport. This would mean that changing airports would force us to "load" again, but that's a trade I would make to be able to load up more quickly.

erikquinn:
^^ Actually, I AM mistaken. 

![image](https://cloud.githubusercontent.com/assets/5103735/13262454/e052e7f0-da32-11e5-93d6-813e53833d86.png)

Apparently, it's actually spending 85.46% of the loading time waiting for the github pages server to start sending the json files.. and that's when it's flying through the load. It's probably way worse when it loads slowly. Thus, we could reduce loading time 100-fold if all the aircraft jsons _were_ merged like @MaicomMR suggests. There are tools for this (like [grunt-minjs](https://www.npmjs.com/package/grunt-minjson)), but having to recompile the minified json every time an update is made to any of the individual jsons would be quite a pain the ass, and arguably not worth it. Maybe that's something that could be done server-side if we do start using something like heroku (with which I have zero experience), and then it would be done automatically and wouldn't be bad at all.

MaicomMR:
I will try to study on grunt-min js & heroku, but I have little knowledge in programming, in fact, I teach history. I'll try to use my spare time to learn more about these programs, my intention was to actually call the members for discussion of this issue.
Very thanks @erikquinn.

headphase:
[...]

PR #561, opened 2016-06-08
Asynchronously load aircraft, airlines and airports

tedrek:
This moves the airline/aircraft load to asynchronous during game play.  A demo is available at http://tedrek.github.io/atc

The tradeoff is that first load of aircraft will be a second or two later than normal, and the player needs to remain connected.  I don't think either of these are a big issue.

This should address #402.

A similiar technique could be applied to airports and geojson files if the airport list information is pre-populated.

### DIFF
--- a/assets/airports/ebbr.json
+++ b/assets/airports/ebbr.json
@@ -1,7 +1,5 @@
 {
   "reference": "https://www.belgocontrol.be/opersite/eaip/eAIP_Main/html/eAIP/EB-AD-2.EBBR-en-GB.html",
-  "name": "Brussels-National &#9983",
-  "level": "easy",
   "radio": {
     "twr": "Brussels Tower",
     "app": "Brussels Approach",
--- a/assets/airports/eddf.json
+++ b/assets/airports/eddf.json
@@ -1,6 +1,4 @@
 {
-  "name": "Frankfurt Airport",
-  "level": "medium",
   "radio": {
     "twr": "Frankfurt Tower",
     "app": "Langen Radar",
--- a/assets/airports/eddh.json
+++ b/assets/airports/eddh.json
@@ -1,6 +1,4 @@
  {
-  "name": "Hamburg Airport",
-  "level": "easy",
   "radio": {
     "twr": "Hamburg Tower",
     "app": "Hamburg Director",
--- a/assets/airports/eddm.json
+++ b/assets/airports/eddm.json
@@ -1,6 +1,4 @@
 {
-  "name": "Franz Josef Strauß International Airport",
-  "level": "beginner",
   "radio": {
     "twr": "München Tower",
     "app": "München Radar",
--- a/assets/airports/eddt.json
+++ b/assets/airports/eddt.json
@@ -1,6 +1,4 @@
  {
-  "name": "Berlin Tegel Airport",
-  "level": "medium",
   "radio": {
     "twr": "Tegel Tower",
     "app": "Berlin Director",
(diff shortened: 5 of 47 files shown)


tedrek:
With these changes the load time on my laptop is now 3 seconds, most of which is due to it being a netbook, so slow parsing.

Airports are now loaded on demand.  However the code, difficulty and name must be specified in airport.js still in order to populate the selector box.

The one thing which it doesn't do and the previous loading system did is retry in case of failure.

Any feedback would be appreciated, especially somebody trying the linked version and saying it does work for them (or things are moving twice as fast as they should be...)

eliuuk:
I do certainly like this better, and loads extremely quicker for me upon opening the webpage - less than a second (excluding the animations)! But, I think what can be worked on is that loading screen between  selecting a different airport, as it doesn't seem right to update the entire page with a loading screen and state what it is loading before actually showing the requested airport.

Overall, it's epic - it does the job quickly and does certainly solve the issue!

tedrek:
Thanks for the feedback, I'm glad it's working for you :smile: (and not just me).

Regarding the loading screen, I agree it's not the greatest (note there is no load screen for switching back to an airport which is already loaded).  Here's the issue, the game needs to pause hard while loading the new airport so the user can't interact with the canvas (this is due to the new airport not yet being in a manipulable state).  When we start the new airport loading we don't know how long it is going to take (geojson files for example can be sizeable).  It could be anywhere from 0.25 seconds where the .json is already cached and just needs to be parsed to 30+ seconds where a large .geojson needs to be downloaded by a user on dial-up.  Unfortunately I don't have any median load time statistics for our users so can't tell you what people are likely to encounter (but I encounter roughly 1 second load times).

What I wanted to avoid is somebody with a long load time sitting there seeing the old canvas but not being able to interact with it leading to the impression the game has crashed.  My feeling at the moment is that it would be best to show an animated spinner if it hasn't loaded after a delay (0.5 seconds?) and keep it up for at least 1 second (flickering screens are generally anxiety inducing).  How does that sound?

eliuuk:
Sounds great! It definitely seems more interesting too! :P

erikquinn:
This is sweet! Works really well for me too. Only thoughts I had were:
- It would be nice if they would continue trying to load all other assets in the background while it's not busy fetching files it needs for a particular operation (such as getting `aal.json` to spawn an a/c with callsign `AAL`). Then delayed spawning would only be experienced for aircraft within the first minute or two, after which everything will be fully loaded. Then we'd really be cookin'!
- small thing: I wasn't really a fan of the lowercase icao codes in the airport selection dialog... perhaps this is a result of the need for case-sensitive file names with `$.getJSON`? If so, maybe a `.toUpperCase()` could be done at the place where the icao codes are written to `$('#airport-list')`. Not a big deal, it just weirds me out to see them in lower case.

Overall, this is great and gets two thumbs up from this guy! :+1: :+1:

tedrek:
@erikquinn, I think this is ready for merge.  I rebased against the latest version and all airports should be present.  We'll need to be careful merging any airports in the near future and make sure they are using the new method of loading.

I've removed the loading screen as my testing showed it under 1 second for most things and I haven't had the time to implement the more complex logic, it can be implemented later without conflicts.

Regarding the display of airport ICAO codes, I didn't change anything in that area so it doesn't make sense as part of this PR.  By all means though, just create another PR for it.

Finally, for further background loading, that's a great idea but I don't have time to implement it so would prefer to merge this now to stop conflicts with new airports.  I'll try to work on the background loading as my next thing.

erikquinn:
@tedrek Cool beans, sounds good! :+1:

PR #563, opened 2016-06-08
Hide some fixes to enhance visibility

Alpi-no:
Those are all more or less unnecessary fixes at crowded areas. Personally I think we should hide all the sid fixes, because they aren't actively used anyway (most the time)


erikquinn:
### Before:

image

### After:

image

---

There's not a big difference here, and honestly, I agree with you when you said:

> Personally I think we should hide all the sid fixes, because they aren't actively used anyway (most the time)

If we're going to clean it up, I'd at least remove all the RNAV fixes that aren't being used by a SID or STAR... unless I'm mistaken, there is no possible value to all of the circled fixes below that could be removed and greatly improve visibility:

image

Alpi-no:
well basically all of them are used in SIDS/STARS, but at least the removal of the SID fixes would enhance visibility by a lot

erikquinn:
@jakobeng1303 Okay, so what do you propose then? Do you just mean the 13 fixes in this pull, because that doesn't seem to me like it would make enough of a difference.

Alpi-no:
@erikquinn I'm not sure if that's on purpose but it seems like there no difference in your before and after shot and that would suggest that it doesn't work. The fixes I have to chosen to change are in positions, where they are likely to be mistaken or blocking important ones (for STARS). 
For some reason you have highlighted some with black circles, however, pretty much none of these are actually affected by the change.

erikquinn:
### Before:

image

### After:

image

@jakobeng1303 These are fresh screenshots. I don't see any difference either, which is kind of my point. If we are going to hide some fixes, I wouldn't remove 13 as you suggested, but I figure we might as well remove ALL (~30+) of the rnav fixes that aren't being used (those were the ones I circled in black). If we hide all of those, until they are needed, that would make it quite a bit better.

Alpi-no:
I'm a little confused now, why are the ones i want to hide still on the screenshots? 
I wouldn't remove any of the 30 fixes you suggested, because they are all part of star. 100% of them. I'm a little curios how you got to believe that they are not needed :D

erikquinn:
@jakobeng1303 AH, I see. I guess they would have to stay then...

And the screenshots are one before the changes are merged, and one after they are merged. So if the fixes aren't disappearing at all, that's a separate problem...

Alpi-no:
@erikquinn I think none of the fixes have disappeared! For example ww230 which is south of Lor 15 Wien should be gone but isn't. I have no clue why

erikquinn:
@jakobeng1303 `WW230` seems to be in use in the `DITIS`, `LANUX`, `LEVDA`, and `MOTIX` SIDs, as do many of the other waypoints (`WW100`, `WW286`, `WW151`, etc), so they can't be renamed unless they are also renamed in the SIDs...

But, if you were willing to forego the majority of the blue lines, we could just change ALL the SID/STAR RNAV points to hidden fixes, which would definitely fix the visibility issues. There is an example of this already in action at KSFO:

image

If we did it that way, then you could decide how much of the SID you want to be pictured (maybe hide everything before the convergence points such as `WW171`, `WW176`, `WW177`, `WW178`, `WW172`, etc. Then all other WW fixes could be hidden from the screen, and they will still follow the right course, it will just say `"[RNAV"]` about where they're flying to during that portion, which makes sense anyway. Or, you could show some of those fixes, but not others... you would know which are important to be able to see or not, and how much to remove. But, I think that is probably the best solution.

PR #567, opened 2016-06-10
Aircraft must takeoff in the order which they taxi - Fixes #565

### LINKED ISSUE #565: Clearing aircraft to takeoff in a different order to taxi causes aircraft to disappear as an object
For example, if I have a few aircraft and taxi them in order:
1. FDB722 A319     X departure
2. UAE222 H/A388  Y departure
3. QTR25  H/B77W Y departure

Then clear UAE222 for takeoff, since it's heavy.

FDB722 will disappear from radar, but will stay on the flight strips and follow any command you give them. Once FDB722 is cleared for takeoff, it will reappear on the radar and takeoff as requested.

tedrek:
This fixes aircraft being able to takeoff before the aircraft in front of them.  It looked to be a piece which wasn't migrated with the overhaul of runways.

@erikquinn, could you confirm this matches the behaviour you expect?

### DIFF
--- a/assets/scripts/aircraft.js
+++ b/assets/scripts/aircraft.js
@@ -114,22 +114,14 @@ zlsa.atc.Conflict = Fiber.extend(function() {
         // If either are in runway queue, remove them from it
         for(var i in airport_get().runways) {
           var rwy = airport_get().runways[i];
+
           // Primary End of Runway
-          var queue = rwy[0].queue;
-          for(var j in queue) {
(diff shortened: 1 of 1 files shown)


erikquinn:
@tedrek Yes. I wasn't aware of #565, I've been out of town about a week. But yes this seems to have fixed it, and your assumption is correct; I must have missed it with that bit of work.

All seems good with this!

PR #574, opened 2016-06-12
[ready & complete] WMPK, Penang International Airport, MYS

eliuuk:
Penang International Airport is a small international airport based in Penang, which is an extremely small island west of Malaysia. Hub for AirAsia, it's also a destination for many Asian international airlines such as SilkAir and Firefly. Malaysia Airlines also run cargo and domestic passenger services to here.

Originally my home town, I now reside in Leeds in the United Kingdom!

### DIFF
--- a/assets/airlines/axm.json
+++ b/assets/airlines/axm.json
@@ -1,0 +1,12 @@
+{
+  "name": "AirAsia",
+  "callsign": {
+    "name": "Red Cap",
+    "length": 3
+  },
+  "fleets": {
+    "default": [
+      ["A320", 81 ]
+    ]
+  }
+}
--- a/assets/airlines/ffm.json
+++ b/assets/airlines/ffm.json
@@ -1,0 +1,13 @@
+{
+  "name": "Firefly",
+  "callsign": {
+    "name": "Firefly",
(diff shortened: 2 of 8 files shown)


erikquinn:
There seems to be something wrong with one of the SIDs, as there is an infrequent crash when the sim tries to look up the current waypoint for an aircraft. Perhaps the `VPG` SID should be removed, as it doesn't seem to do much. But I'm not sure what is causing this issue. I sometimes get a freeze when issuing `caf cvs w` to the departures as I always do when testing the SIDs.

eliuuk:
@erikquinn I'll do more testing on this tonight and have a look, thanks!

eliuuk:
I've done a little bit of testing just now, and I can't seem to reproduce this, but I've managed to somehow get an aircraft circling around VPG

erikquinn:
Yeah, I've tried some more and can't get it again. It was hard to make happen, so isn't real concerning, and it may have _nothing_ to do with this pull either; it could be something in the code that happens at every airport under a very specific set of circumstances. I _did_ change the wind so it wasn't at a 90 degree angle though, which may have something to do with it.

PR #581, opened 2016-06-18
Improve logic for checking when to proceed to next fix

### LINKED ISSUE #456: "loop-dee-loop" of confusion
Often, aircraft will be projected to do big loops around fixes they're navigating to, presumably because they are projected to "miss" the fix slightly, and then try to loop back around to cross directly over it like they were trying to. In reality, passing right over the fix is not super important. I'm not sure _how_ the logic works to determine whether you are close enough, but for our purposes it could be as simple as increasing the allowable lateral deviation from the fix's position to consider the aircraft "over it" sufficiently to proceed to the next waypoint.

image
In this screenshot, the aircraft ends up slightly east of `ANKER`, and therefore turns hard left to try to reach it.

erikquinn:
Especially on a slow computer or at higher simulation rates, it is noticeable that aircraft miss their waypoints slightly, because the distance covered between two position recalculations is outside the radius acceptable to consider the aircraft to be "over the fix". Instead, it now checks whether they have _PASSED_ the fix, thus resolving the issue.

This pull resolves #456.


tedrek:
I think this has a bug, if you order an aircraft to a fix which would require a greater than 180 degree turn then it will immediately proceed to the next fix, even if the requested fix is 100km away.  I'm not sure the best way of fixing that, perhaps keeping the previous offset and switching to the next fix when the offset goes from positive to negative?

erikquinn:
@tedrek Yeah, you're right. I thought about this for a _long_ while today, and can't come up with a good solution, sadly... Best I can seem to do is to limit it so it will only move on to the next waypoint if it has passed it and is within a certain distance of the fix. Then, worst case scenario, it would skip the waypoint that's close behind them, and then enter the turn to the second one... it's not ideal, but would at least solve the main issue at hand I guess.

tedrek:
An unstated assumption here is that the game should behave well as the simulation step size increases without bound.  However that's probably a bad approach to take.

assets/scripts/module.js:304 limits the step size to 1/20 of a second.  With 5x warp the step size can increase to 1/4 of a second.  My thinking at this point is the step size should be limited to 1/5 of a second as the value at which two aircraft will reliably collide when in a head-on course at speed and altitude.  This means an aircraft would not move more than roughly 50 meters in a single step.

#456 is not caused by a slow computer, rather it's caused by the step time being increased to 5 seconds during projection.  This means aircraft can be moving up to 1.25 km per step which makes it pretty clear why the 1km distance doesn't work reliably for projections.

erikquinn:
@tedrek okay, understood. I guess part of why I went this seemingly strange route is because I like to let the game "run for hours" as a lazy last step of testing something (particularly new airports). I fling the simulation rate up to about 1000x and in less than 5 minutes real time, it's pretty well guaranteed to have spawned all combinations of aircraft, and then I know there aren't any problems with airlines, fleets, SIDs (assuming I depart a few), STARs, etc. It's caught more than a few things I missed when reviewing others' work, and the spinning kind of just drove me bonkers, that's all.

It actually boils down to two things when at high sim rates:
1. The projection lines aren't always accurate, because the step size can become enormous.
2. The aircraft might not realize when they "pass" a fix because they hopped _over_ the point.

---

Paths forward that come to mind:
1. set a cap of ~300m maximum projection step size
   - :+1: resolves item 1
   - :x: limits the allowable maximum simulation rate
2. use changed logic for passing fixes, as proposed in this pr
   - :+1: resolves item 1
   - :+1: resolves item 2
   - :+1: does not limit allowable maximum simulation rate
   - :x: when <~5nm from the current fix, facing away, it will skip over that fix (undesirable)
3. spend fewer cycles on projection and more on positional precision
   - :+1: resolves item 1 because more data will be available
   - :+1: resolves item 2 because more data will be available
   - :+1: end up with a higher density of positional data
   - :ok_hand: would have to run some timers that call `canvas_draw_aircraft()` at a specifyable interval in order to update at a different rate than the position recalculations (but that doesn't worry me)
4. do not attempt to conduct projections above 50x simulation
   - :+1: resolves item 1
   - :x: does nothing to address aircraft who actually do miss their fixes (item 2)
   - :x: no projection lines, obviously

Way I see it, I'd personally rank the options 3-2-1-4.

How do you feel about these? Or any other ideas you may have?

Thanks.

erikquinn:
Option 3 is the plan, if projection isn't just to be removed altogether. Regardless, this can be closed for now.

PR #590, opened 2016-06-25
[ready & complete] VABB, Mumbai Chhatrapati Shivaji International Airport, IND

eliuuk:
Chhatrapati Shivaji International Airport is the airport of Mumbai, India. It is the second busiest international airport in India after Delhi, and is also a hub for Air India, Air India Regional, Blue Dart Aviation, GoAir, Jet Airways and JetKonnect. Additionally, it's a focus city for IndiGo.

Mumbai has also won a few awards, but most notably is Best Airport in India and Central Asia, Skytrax World Airport Awards 2016.

Public testing: https://indianbhaji.github.io/atc/b/general/


erikquinn:
This is about a 500nm diameter... is that really how much the _approach control_ facility owns? Seems like that would also include surrounding enroute/center airspace. I guess that's not inherently a problem, but the scope of the game has always been approach/departure services, and this thing would be a constant battle of zooming in/out.

@indianbhaji Thoughts?

image

eliuuk:
@erikquinn Well, I can tweak it slightly according to the Indian AIP (which is really outdated!)

http://www.aai.aero/public_notices/aaisite_test/eAIP/PUB/2012-04-01/html/index-en-GB.html

The coordinates of the Mumbai TCA are provided in that AIP under "Mumbai Control Area".

But then, here's a screenshot of the newer version:
http://prntscr.com/bvrn2i

erikquinn:
@indianbhaji Huh. It looks like that must really be the right size... I just find it very strange that the approach control facility has such a remarkably enormous area of jurisdiction. From what I've seen, nowhere else on the planet comes even close. For example, there's well over four times as much traffic in New York TRACON (N90), and yet Mumbai approach is about 6 times larger than N90. Just odd.

I guess we can deal with it. It may make a helpful testing bed for airways and enroute-mode stuff. Whatever the most accurate boundaries you have then I guess will work!

I'll get back to checking this out later.

erikquinn:
image

To explore another option here... What if, instead of the whole airspace, we only use the _inner portion_ within the 100nm ring depicted? I assume this radar facility has to be split into multiple sectors worked by multiple controllers; perhaps we can just simulate the controller working this portion? It's unclear what the blue dashed lines mean, as they are not in the key of this document I found, but would this be a reasonable compromise, or do you think it should remain the size of the entire facility?

eliuuk:
In this case, I may have to redo the SIDs as I have expanded some of them 
to allow spawning outside the airspace that I've currently developed. I'll 
be looking to minimise the airspace size when I'm back on Friday.

eliuuk:
@erikquinn I'm back, finally (after sorting out a lot of stuff)! So, for Mumbai, are you proposing we only include the yellow ring as part of the airspace?

erikquinn:
@indianbhaji I was thinking about it... does that sound reasonable to you? We obviously want to be realistic, but that airspace is so huge that I'd be inclined to just use the inner ring, unless that seems dumb. If it seems dumb to you, we can stick with the full size, I'm just exploring the options.

eliuuk:
I see no reason as to why your suggestion is 'dumb', so I'll revise the 
airspace tomorrow.

erikquinn:
Alrighty! It'll still be one of the largest ones we have, but I think this will be a decent compromise between realism and playability. :smile:

eliuuk:
@erikquinn That should be it! I've tried terrain with this airport too, but my computer just can't handle the part with the natural neighbouring so, unfortunately, no terrain until I upgrade my computer.

PR #594, opened 2016-06-28
[ready & complete] VOBL, Kempegowda International Airport Bengaluru

eliuuk:
Kempegowda International Airport Bengaluru is an international airport, which serves the Indian city, Bangalore. Based here are 4 airlines, and Kempegowda is also a focus city for 2 airlines.  This airport served 18 million passengers throughout 2015 and 400 aircraft movements per day in 2015. 

This airport opened in 2008 as an alternative to HAL Bangalore Airport, but now is the main airport in Bangalore.

The airport is named after Kempe Gowda I, who was the founder of Bangalore.

Public testing: https://indianbhaji.github.io/atc/b/general/


erikquinn:
Same comments about the size as I had in #590. Will wait on this PR until the discussion is resolved on #590 so we have a plan on what to do with supermassive airspaces like this.

image

image

eliuuk:
@erikquinn For this aerodrome, would the 90nm ring make sense?

erikquinn:
@indianbhaji I don't know... the shelving they've got going on makes no sense to me anyway. Doesn't look like there's much of an easy and logical way to take a subset of the depicted airspace.

What do you think is the best thing to do with it?

eliuuk:
More scavanging. I'll see if I can find anything off the vACC website on 
VATSIM or IVAO.

eliuuk:
@erikquinn IVAO uses a 50nm approach circle around the fix BBG. This should now be fixed!

eliuuk:
@erikquinn Any chance this could be merged after I fix the conflicts tomorrow?

erikquinn:
Perhaps! Please make sure the PR is working against the current version of the repo though, and not an old one, as the differences are pretty drastic from how the codebase was set up a few weeks ago.

assets/scripts/airport.js
```diff
@@ -1346,6 +1346,7 @@ function airport_init() {
   airport_load('sbgr', "beginner", "Aeroporto Internacional de São Paulo/Guarulhos");
   airport_load('tncm', "easy", "Princess Juliana International Airport");
   airport_load('uudd', "easy", "Moscow Domodedovo Airport");
+  airport_load('vobl', "hard", "Kempegowda International Airport Bengaluru");
```
n8rzz: in `v3`, this line will need to go in the `airportLoadList.js` file. see the comments in https://github.com/zlsa/atc/blob/gh-pages/src/assets/scripts/airport/airportLoadList.js for more information here.

We have #719 and n8rzz#88 which will eliminate the need for any npm commands when adding new airports/airlines/aircraft.

eliuuk: @n8rzz I'm rebasing any remaining pull requests I have anyways.

I'm unaware of what you mean by 'npm commands', all I know about loading an aerodrome is that you have to add it in with the new format in the file `bundle.js`. Could you explain what you mean?

n8rzz: @indianbhaji the `bundle.js` file is actually a generated file.  It contains all of the js for the app in one file and isn't meant to be edited directly.  Now, you _could_ edit it directly, but be aware that the next time a build is run, those changes will be overwritten.  A good work around, until #712 is in place, is adding the new airport to both the `bundle.js` and `airportLoadList.js`.

What is npm?  NPM is short for Node Package Manager.  Sort of like bundler for Ruby.  It provides a means to manage packages and run scripts with those packages.  v3 is written in ES2015 (or ES6 as its also known).  Browsers don't actually understand ES2105 thus, it must be transpiled to ES5.  It gets complicated.

Long story short, when developing on the JS one will need to run `npm run build` in order to re-build the js.  When developing airports, this will also be a necessary step until #712 is completed.  A temporary workaround would be what I mentioned above.  It is very important that those new airports get into the `airportLoadList.js` otherwise they will not load after the next build.

eliuuk: @n8rzz I understand now, thanks for your explanation! +1

erikquinn:
@indianbhaji I think I can start dredging my way through all the PRs that have backed up. Is this one still fully ready?

eliuuk:
@erikquinn Looks like it's fully ready, you can start looking at it!

erikquinn:
Merged via d3143fe1d81f50d1318c12ddb330d5154fd9cfd2.

PR #583, opened 2016-06-19
New Airport: SAWH (Ushuaia Airport)

Fechulo:
Ushuaia Airport, officially "Aeropuerto Internacional Malvinas Argentinas" (A very controversial name as it reflects Argentina's claims of sovereignty over the Falkland Islands) is a small airport at the southern tip of Argentina in Ushuaia, the southernmost city in the world. I made this airport mainly because I wanted to play around with terrain (I think it looks pretty good, although there are some errors north of the airport, where there's two overlapping areas with an elevation of 1000ft).

This PR also includes changes to LAN and TAM, since the 2 airlines have merged into one airline called LATAM. 

When #561 is merged, d8d0de9 and c9df00e will become unnecessary.

PR #604, opened 2016-07-05
New Airport: SAWH (Ushuaia Airport)

Fechulo:
Ushuaia Airport, officially "Aeropuerto Internacional Malvinas Argentinas" (A very controversial name as it reflects Argentina's claims of sovereignty over the Falkland Islands) is a small airport at the southern tip of Argentina in Ushuaia, the southernmost city in the world. I made this airport mainly because I wanted to play around with terrain (I think it looks pretty good, although there are some errors north of the airport, where there's two overlapping areas with an elevation of 1000ft).

This PR also includes changes to LAN and TAM, since the 2 airlines have merged into one airline called LATAM.

It fixes the issues discussed in #583


assets/airports/sawh.json
```diff
@@ -0,0 +1,149 @@
...
+    "PWL3": {
+      "icao": "PWL3",
+      "name": "Williams Three",
+      "rwy": {
+        "25" : [["_RWY25", "A4+|S180"], ["WH321", "A17+|S180"], ["WH322", "A50+|S180"], ["WH323", "A75+|S180"], ["WH324", "A80+|S180"]]
+      },
+      "transitions": {
+        "USU": [["WH326", "A110+|S180"], ["USU", "S180"], "UGIRA", "LOKEN"],
+        "USU": [["WH326", "A110+|S180"], ["USU", "S180"], "KEXOP", "GRA"],
+        "USU": [["WH326", "A110+|S180"], ["USU", "S180"], "LITOK", "ALMIR"],
```
tedrek: Only this USU transition is used, it overwrites the previous ones.  I don't think that's the intent, they should probably be renamed based on the last fix.

tedrek:
That terrain is pretty crazy :+1: 

I was having difficulties with the USU2B.GRA and PWL3.DABLI departures interfering with the DABLI2 arrival.  Is that the an expected challenge for this airport?

Fechulo:
regarding the USU transitions, I'm trying to simulate airways. The SID actually ends at USU, but aircraft wouldn't just continue on that heading, they would follow an airway. I'm aware that airways can be added, but I don think they're functional yet (not to my knowledge). If there's a better way to do this, let me know!

I did a bit of research and it seems like, no matter where aircraft are coming from, they'll all end up at DABLI, which is probably why DABLI2 is the only STAR at the airport. Pretty much the same thing happens with departures. Most aircraft departing the airport must, at some point, end up at GRA, because there's many airways which connect the airport to the rest of the country. I imagine this is an issue ATC IRL must deal with every day. When I tested the airport, there weren't any conflicts, because there was plenty of vertical separation. To be honest, I don't really know what to do about this.

tedrek:
@Fechulo, given that airways aren't yet implemented I think using the transitions on USU is a good solution.  I was commenting on the technical aspect, you can't use the same transition name multiple times.  `{"foo": "bar", "foo": "baz"}` becomes `{"foo": "baz"}` when it is parsed.

My suggestion would be to change it like this

```
  "_comment": "Simulate airways after USU using transitions",
  "transitions": {
    "LOKEN": ...,
    "GRA": ...,
    "ALMIR": ...
  },
```

Regarding the vertical separation between GRA and DABLI, I am fine leaving it as is.  Within the control area the player can resolve issues and it makes for interesting game play.  Outside of the area I was seeing conflicts and potential conflicts but I can live with it.

One potential solution though would be to modify the STAR so they travel a parallel path a little further to the east  while outside of the control area.  When they enter the control area do a little jog onto the correct path.  It maintains the challenge for the player without causing issues they can't control.

erikquinn:
Just a comment: The framework for the airways is in there, you just have to make sure the route the ac is on has that in the flight plan (eg SAWH USU2B USU V123 ABCDE V345 FGHIJ). If the route can be set to that in the departures flight plan (I don't recall how it's set up), then the fms will at least know how to follow the airways.

Fechulo:
@erikquinn 's solution seems more elegant. I'll change it as soon as I know how to do it. @tedrek I'm going to try and see how ATC IRL deal with this issue, before I make any changes.

Fechulo:
@tedrek Upon further research, I found that controllers in real life, don't deal with this issue at all. There's simply not enough traffic at the airport for there to be any conflicts.

Fechulo:
@erikquinn Weren't there airways at KSFO? I can't find them anywhere in the code. Are there any examples of airways in the game?

tedrek:
@Fechulo, I've taken a look at the code regarding airways.  There isn't currently an ability to assign a route to departures, just a single SID.  I will try to work on that, but it's going to take a couple of days at minimum.

Fechulo:
That's fine, no problem. Thanks!

erikquinn:
@Fechulo imma slap together some airways for this and give it a try. Could someone direct me to some charts for SAWH though? I had trouble finding any SIDs/STARs via the typical google efforts...

erikquinn:
PS the airways implementation is straightforward, as demonstrated [here](https://github.com/zlsa/atc/commit/1754306fa86e322aac4a5dd77af976ebfb45753c#diff-0112be1bce5714389b11f0fcf314a677L114). I swore I documented it thoroughly, but must have messed something up in the process of cleaning up my work and creating the pr, merging etc, and it got lost along the way... :disappointed:

Regardless, I see the issue now is that you want to add an airways leg as a part of a SID, which won't work well with how it's set up... both the SID and the airway are (in the eyes of the code anyway) procedures that have an entry point, exit point, and a bunch of fixes inbetween. It wouldn't work to have `entryfix.airway.exitfix` stuck in the middle of `entryfix.SID.exitfix` as implemented now; that would have to be an additional change.

So, I would probably say that what you did with just adding the fixes as a list of fixes on the transition is the way to do it.

Fechulo:
The best charts I could find while making the airport were [these](https://drive.google.com/file/d/0BxKQrDika6dOdUQ0bmJTYUxueVk/view) ones. I don't know if they're new or old, because they don't have a date, but they're the best I could find online. If you actually want to get the real charts you have to go to a physical [shop](https://www.google.ie/maps/place/Av.+Paseo+Col%C3%B3n+1452,+C1063ADO+CABA,+Argentina/@-34.6246993,-58.3680373,3a,41.7y,78.83h,90.94t/data=!3m7!1e1!3m5!1sVVcxWyiumcfqN9QnyIAoHg!2e0!6s%2F%2Fgeo0.ggpht.com%2Fcbk%3Fpanoid%3DVVcxWyiumcfqN9QnyIAoHg%26output%3Dthumbnail%26cb_client%3Dmaps_sv.tactile.gps%26thumb%3D2%26w%3D203%26h%3D100%26yaw%3D47.101604%26pitch%3D0!7i13312!8i6656!4m5!3m4!1s0x95a334cd97e07b8d:0x42276cc7b06e122c!8m2!3d-34.6246626!4d-58.3680392?hl=en) in Buenos Aires to buy them.

As you can see in the charts, the airways are not actually part of the SID, but I wanted the a/c to leave the airspace as realistically as possible. I added the airways that they could follow from the USU VOR to the SID as if they were in the procedure.

I could swear that there were airways implemented at KSFO, but when I looked for them, I couldn't find them. I guess I'll rename the transitions so that the SID actually works.

willfrd:
I could give you really recent charts if you wish. But I had a look at them and they haven't really changed. Do you want them?

Fechulo:
Thanks for the offer! If they haven't changed, I don't think I need them. The airport is good as it is now, and honestly I have no intention on changing anything, unless it's necessary. Thank you anyway!

willfrd:
Ok you are welcome :-) Just remember if you or anyone here need any recent charts ask me and I'll upload them to drive (If I have them which 99.9% of the times I do), I don't think this should be an obstacle when it comes the time to write up an airport.

Fechulo:
@willfrd I assume you have navigraph charts. In that case, I'm not sure they would be happy with someone uploading their charts to the internet without their permission.

willfrd:
Not navigraph but jeppesen... And as I have the license... I don't think they will mind but if they do don't worry its my problem :-) ahh and I would upload them temporarily or send it privately ;-)

erikquinn:
> It wouldn't work to have entryfix.airway.exitfix stuck in the middle of entryfix.SID.exitfix as implemented now [...] So, I would probably say that what you did with just adding the fixes as a list of fixes on the transition is the way to do it.

@Fechulo are you okay to just move forward with this as it is? Enabling nested procedures (eg an "airway" inside a SID/STAR) is not something that's coming from me any time soon. Being able to define a ROUTE to fly instead of just a SID could be added, but also is not right around the corner, so let me know if you want to continue with this.

Fechulo:
I'd rather have it merged now. It's not my priority either and it's functional as it is right now. Also, it's better to merge it now before it becomes outdated, and so that #622 can be merged as well.

PR #605, opened 2016-07-07
Added delete command

glott:
Added `delete` command. Can be used to remove aircraft from the scope. 
Syntax: [Select aircraft] `delete`
Synonyms: `del`, `rem`


tedrek:
I'm curious what is the intended use case for this?  It is certainly not very realistic :wink:

erikquinn:
It's present in all the major simulators (NexSim, Adacel, UFA, etc), and I can attest there are times I've wanted it... @tedrek it's definitely not realistic, but it's important from the training perspective, as it allows you to focus on a particular task (say box pattern vectoring), and if you mess up, you have the option to kill a plane or two and continue with the drill. Useful for developing skills, and personally i would rather have the option to use or not use it based on whatever purposes I'm playing the game that day.

assets/scripts/aircraft.js
```diff
@@ -2151,6 +2156,11 @@ var Aircraft=Fiber.extend(function() {
 
       return ["ok", "proceed direct " + fixname];
     },
+    runDelete: function(data) {
+      for(var i=0;i<prop.aircraft.list.length;i++)
+ +      if(prop.aircraft.list[i].getCallsign() == this.getCallsign()) prop.aircraft.list[i].to_remove = true;
```
tedrek: @JHG0, these  2 lines should be simplified to `this.to_remove = true`, the command dispatch logic has already verified no other aircraft match this callsign.  After that I am ready to merge.

tedrek:
@erikquinn, thanks for the insight.  I'll merge once my comment on the diff is resolved.

glott:
I'll fix it when I have a chance. What do you think would be the best/most efficient way to give the option of deleting an aircraft when they are not in the center airspace? @erikquinn @tedrek

tedrek:
@JHG0, the best solution is probably to wait for #601 to be merged and then implement it as a system level command which will delete regardless of being in the airspace.

glott:
Closed - new pull request will be made after merge with #601. @tedrek

PR #610, opened 2016-07-09
Better Traffic Mix For Dublin

EvilDoctorK:
requires stk.json - see my other PR , sorry not very good at github so can't seem to merge the two PRs

### DIFF
--- a/assets/airports/eidw.json
+++ b/assets/airports/eidw.json
@@ -1009,13 +1009,16 @@
     "airlines": [
       ["ein", 15],
       ["ryr", 10],
+      ["stk",  7],
       ["baw",  5],
-      ["dlh",  5],
-      ["aal",  3],
-      ["aca",  3],
-      ["dal",  3],
-      ["ual",  3],
-      ["klm",  1]
+      ["bcy",  5],
+      ["bee",  3],
+      ["aal/long",  3],
+      ["aca/long",  2],
+      ["dal/long",  3],
+      ["ual/long",  3],
+      ["swr",  1],
+      ["ibe",  1]
     ],
     "destinations": [
       "SUROX", "LIFFY", "BEPAN", "INKUR", "DEXEN", "OLONO", "ROTEV", "NEVRI", "PESIT"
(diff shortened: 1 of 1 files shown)


Fechulo:
I've been wanting to update the traffic at EIDW for a while, but I was too busy, so thanks :smile:! The addition of Aer Lingus Regional brings a tiny problem: If you're familiar with the procedures at Dublin, you'll know that there's different SID's for CAT A/B (e.g. AT72) aircraft and CAT C/D (B747) aircraft. The SID's in the game are only the CAT C/D SID's. Adding the other ones is pretty easy, but there's no way of telling the game that those SID's are for CAT A/B aircraft only.

EvilDoctorK:
True enough, though that would be beyond my coding ability to fix that !

Fechulo:
I wasn't expecting you to do it, I can't even do it myself. I was just pointing out the issue so that someone, like @erikquinn or @tedrek can maybe consider this.

erikquinn:
Realism is paramount, I would say, but it sounds like the consensus is that there are improvements in here that would be good to merge... @Fechulo what would you suggest regarding the new airlines operating cat A/B aircraft? Leave them in there and add the other procedures, and allow any a/c type to fly any A/B or C/D SID? Or would it be better to remove those small-aircraft-operating airlines from this, and merge the rest?

I can make the change during the merge process, but am not sure what the best resolution would be as I think I've played Dublin like _maybe_ two or three times, and only because I was testing stuff. haha

Fechulo:
I really think the airlines should be merged as they are right now, I'm not sure about the procedures though.  On one side, adding more procedures would be better for obvious reasons. On the other hand, C/D aircraft using the A/B procedures would be extremely unrealistic. In the case of EDDM, there are some procedures which are 'propeller-only procedures', but propeller a/c could still fly normal SID's, so I just left those SID's out. To be honest, I don't know what the best plan of action is...

eliuuk:
I included the prop departures in Manchester, which I believe are the LISTO 
departures for the 23s. So, what do we do about these SID differences?

Fechulo:
@indianbhaji In Manchester prop aircraft can also fly the other SID's, while in Dublin, prop a/c can only fly the A/B departures and jets can only fly the C/D SID's.

erikquinn:
@fechulo FAA does the same thing in the notes on each procedure, often citing "Turbojet aircraft only" or "Turboprop/Piston aircraft only"... Therefore, seems like a reasonable thing to add, and wouldn't be hard. Could you write up an issue for it so it can eventually be added as a feature?

Fechulo:
[Here!](https://github.com/zlsa/atc/issues/618)

PR #616, opened 2016-07-14
Support specifying airport and runway elevations

tedrek:
@erikquinn, @Fechulo, here's an implementation for having an airport at some elevation other than 0.  If a runway doesn't specify an elevation then the elevation of the airport itself will be used.


Fechulo:
Thanks!! The conflict detection code has to be changed as well; aircraft in the ground at SAME show a conflict, because they're "not on the ground"
conflict

tedrek:
@Fechulo, thanks it should be fixed now, can you try again?

Fechulo:
Yeah, it works perfectly now! Ready to merge?

assets/airports/ksfo.json
```diff
@@ -127,27 +127,27 @@
     {
       "name":        ["28R", "10L"],
       "name_offset": [[0, 0.3], [0, 0.3]],
-      "end":    [ ["N37.6135324", "W122.3571410"], ["N37.6287386", "W122.3933911"]],
+      "end":    [ ["N37.6135324", "W122.3571410", "13.0ft"], ["N37.6287386", "W122.3933911", "5.5ft"]],
```
erikquinn, line 130: @tedrek Just curious... where are you getting these to-the-inch elevation values? A DEM model maybe? Seems super high precision.

erikquinn:
@tedrek This seems a solid solution, though I feel like I'd be remiss not to bring this up:
Especially if an enroute mode comes along, there will be many, many different ways we'll need to measure both "altitude" and "speed". What I was thinking (I may have mentioned it recently, not sure) was that the best way to deal with it would be to:
- make `aircraft.altitude` into `aircraft.alt`, which has properties `.msl`, `.agl`, `.fl` (for altimeter-adjusted altitude, radar altitude above terrain, and non-adjusted std-altimeter-settings altitude)
  - makes easy to access altitude for assignments and crossing restrictions, terrain/obstruction clearance, and would natively recognize separation losses when attempting to utilize an unusable flight level (eg FL180@A2988)
- make `aircraft.speed` into `aircraft.spd`, which has properties `.ias`, `.gs`, `.mach` (for indicated airspeed, aka the unit we speak to each other about, groundspeed aka speed the dot moves across the ground, mach aka speed relative to sound)
  - makes easy to get indicated speed, and mach speed, which don't have a linear relationship, which will become important if aircraft are to be simulated at altitudes around FL300 or higher

So this pull seems to work as written, but the necessity of remembering to repeatedly add/subtract the field elevation makes it harder in the long run I would think, so I figured I'd bring that idea up. Of course, this all isn't to say we shouldn't just merge this and do the rest later, it's just something to chew on! haha :smile:

assets/airports/ksfo.json
tedrek, line 130: @erikquinn, I pulled them from https://www.airnav.com/airport/KSFO which indicates the field elevation is surveyed and I assumed the runway elevations are too.

erikquinn, line 130: @tedrek huh, I didn't realize airnav posted it to the tenth of a foot! guess I just never looked that closely. Good to know. Thanks!

tedrek:
@Fechulo, yes I am going to merge.

@erikquinn, the add/subtracts are definitely an issue.  My preference would be to implement most of it as methods: `altitudeAGL(), altitudeMSL(), etc.`. Probably `altitude()` should be the same as `altitudeMSL()`.  My reasoning is that would allow changing which ones are calculated on demand and which ones are cached without breaking other parts of the program.  The same principle applies for the speed values.

As an aside, my strong preference is to use whole words for variables/properties/methods.  I definitely read the code many more times than I write and it is much easier for me when full words are used.

erikquinn:
@tedrek makes sense, sounds like a fine plan to me!

Issue #569, opened 2016-06-10
Airport development log


jpeterson1346:
Perhaps for everyone's reference, there should be a log for expected/working/partially completed/etc. airports to prevent duplication of work. A page/post/working document would also provide non-developers insight into what airports can be expected. I'm bringing this up because there are several major airports that lack coverage within the simulation (e.g., my current WIP, KDFW) and consequently, a non-zero potential for users to put unnecessary work into already working projects. Furthermore, it introduces the possibility for developers to request and/or present feedback and other helpful information from others, particularly for developers in similar regions (i.e., people familiar with FAA/NWS practices vs. those with European services).

Another option would be an option for (a) collaboration tool(s) on airspace design, though that introduces a multitude of potential problems.

For substance, perhaps a spreadsheet with the following information might be helpful:
- airport;
- working status;
- expected working status (i.e., will you keep working on it? Is it on hold? etc.);
- complexity (of the airspace);
- estimated completion stage (subjective);
- comments.

Elaborations, suggestions, comments, the like--all welcome.

eliuuk:
This is generally possible just through the use of a pull request, including a checklist on what you are working on in the description and update the title when needed to state the current status, since files automatically update when you push them to your own fork.

jpeterson1346:
You are right on this however my main concern is over the duplication of work. With the amount of current contributors and potential ones, duplication of work seems like a growing possibility. Furthermore, due to the complexity and relative tediousness of this work, redundant work can be rather problematic.

eliuuk:
Duplication is an issue, and I believe that's happened once with KMIA, however, I have no thoughts of any solution to this.

Alpi-no:
Personally, I don't think duplication of work is that much of an issues currently, however, it would be solved by creating an issue and being assigned to it. That would also create a little forum for contributors of the same airport

erikquinn:
@jpeterson1346 I agree it would be nice to see what people are working on, and have some way to collaborate more WHILE doing the work, check in on others' progress, see what is "next on the list", etc... 

There are things out there that can achieve this that I suppose we could consider. For example, something as simple as [slack](https://slack.com/) (common among developers) could address the collaboration portion in a simple way, and has tons of integrations for things like [appear.in](https://appear.in/) where you can have live meetings and web chats, etc.

[...]

PR #617, opened 2016-07-15
Add KIAD (my new home airport!)

erikquinn:
image

Next week sometime, I'll start my new job in the ramp tower at Washington-Dulles Int'l here in northern Virginia... and since I already showed them this simulator, I feel like I will be chastised for not adding Dulles, so here we are! :laughing:

Big airport, lots of planes. Just like I like it.

PS: For what it's worth, the airspace boundary is totally made-up on the top side because I'm lazy and didn't want to reformat Potomac TRACON's whole Shenandoah area video map... so I just used the Mt Vernon area VM (from DCA) instead. The real one extends much, much further north, and a bit northeast as well. Oh well!

PPS: It feels almost Atlanta-like with them triple parallels... :wink:


willfrd:
Nice airport and congrats for your new job @erikquinn !

erikquinn:
thanks! :smile:

Fechulo:
Congratulations!!!! I was actually just about to start working on KIAD, so doing something about #569 is more necessary than I thought.

PR #619, opened 2016-07-18
Added elevations & fixed SIDs

eliuuk:
Added elevations for all airports, excluding American airports.

SIDs fixed in Abu Dhabi and Manchester.

A few other fixes also included, such as removal of some duplicate waypoints.


erikquinn:
Merged via 2f25b2d.

PR #622, opened 2016-07-19
New Airport: SAME (Mendoza Airport)

Fechulo:
More terrain testing. I chose this airport, because the highest mountain outside Asia is just northwest of the airfield. **This PR requires #604 to be merged**


erikquinn:
Depends on #604.

PR #629, opened 2016-07-30
New Airport: KPDX (Portland International Airport)

Fechulo:
This is my second take at an American airport, and to be honest, I'm pretty proud of it. This is also the first time I make an airport with a video map! There's some interesting mountains to the east of the airport, but unfortunately, the P80 TRACON is not big enough to see them.
kpdx


erikquinn:
lookin' good! workin' good! :+1: :+1:

erikquinn:
@Fechulo PS don't throw away the qGIS files, as we may want to come back later and clean up the edges a bit. Overall, very nice work though! It's a fun little airport I think!

Fechulo:
Thanks for the feedback! I was surprised as to how easy it actually was to add the video map. Nowhere near as difficult as terrain! :smile:

erikquinn:
True dat... The terrain is by far the most painful... lol

Issue #638, opened 2016-08-13
A few more commands


erikquinn:
A command should be added for the following things:
- to delete aircraft from the simulation ("del" / "kill")
- have an aircraft fly present heading ("fph")
  - useful to have an aircraft stop flying by fixes, and simply level the wings on their current heading, to expect further vectors.

erikquinn:
Fixed in #641. Closing.

Issue #621, opened 2016-07-19
Land shortkey * no longer functional


eliuuk:
I'm unsure if I'm doing this wrong, but here we go:

I do:
QTR508 \* 16R
= Command not understood

The shortkey does not seem to be working anymore for land, but the other aliases do still work.

erikquinn:
Fixed in #641. Closing.

PR #641, opened 2016-08-14
Fix asterisk shortkey (land) and add del/fph commands

### LINKED ISSUE #620: Commands are case sensitive
It appears that commands are case sensitive;
ie. QTR508 I 16R
= Command not understood

QTR508 FH340
= Command not understood

but...
QTR508 i 16R
= QTR508 cleared ILS 16R approach

QTR508 fh340
= QTR508 fly heading 340 degrees

erikquinn:
Resolves #620, #621, #638.

Adds two new commands, `fph` for "fly present heading" and `kill / del / delete` to remove an aircraft from the simulation. Also removes commands' case sensitivity and returns to service the asterisk shortkey for approach clearances.


Fechulo:
Works perfectly, I'll definitely use the fph all the time!

PR #642, opened 2016-08-16
New Airport: TJSJ [Luis Muñoz Marín Intl. (As requested by @Exfil)]

### LINKED ISSUE #637: New Airport Request: TJSJ
A request was made for Luis Muñoz Marin (San Juan, PR) by @exfil in #634. @Fechulo said he was on it! Creating this issue so it doesn't get forgotten when I merge and close #634.

Fechulo:
Unfortunately no video map for this one, but it still has terrain!

Resolves #637.


erikquinn:
> @Exfil The airport's finished! While it's being reviewed, you can test it [here](https://fechulo.github.io/). I hope you like it!

Exfil:
Thanks to everyone who worked on this! So far both test versions work good and my record so far is aprox 1200 points :)

Fechulo:
@Exfil Thank you for playing. While working on the airport, I wasn't sure what the default runway for arrivals should be. On one hand, runway 8 is the longest, but the extended centerline is very close to the airspace border. runway 10, on the other hand, is shorter and it doesn't have many waypoints on the ILS. I tried both variations and didn't really see a difference in terms of difficulty. What do you think? I imagine you know more about San Juan than I do. If that's the case, do you know if there's a preferential runway system?

Exfil:
Ha yes I live near SJU and the runway they use the most for landings is 10 because most flights come from the North West so its easier to line them up on RWY10. Most takeoffs happen on runway 8. Its not ALWAYS the same runways for takeoff landing so sometimes landings can happen on 08 and takeoffs on 10. Inverse operations only happens like twice a year for a few moments due to wind change. I don't work in ATC so I can't tell you the way they determine how planes land but I know a bit based on what I see regularly.

Fechulo:
Great, so it's good as it is now, thanks for your help!

Exfil:
If I may ask, how does the review process to put the airport in the real game work? How long does it normally take to get approved? I'm not desperate or anything just curious. I play the game everyday on the test link provided and enjoy it a lot.

Fechulo:
Well, it depends on how much free time collaborators have, so it can take anything from a day up to several weeks.

erikquinn:
@Exfil Sorry for the delay... I'm having a look now.

@Fechulo If you don't mind me asking, what was your source for the airspace boundary?
image

erikquinn:
I did some googling and can't really get any good data for it, but it looks like the full boundary looks something like this:

image

Also, ZSU (like the HCF in Hawaii) is a "CERAP" (combined **CE**nter & **R**adar **A**pproach **C**ontrol), and thus the large airspace is not really divided between a "center" area and an "approach control" area... So we would obviously run into the same problems @indianbhaji and I discussed for some malaysian airports; @Fechulo, was the included airspace your solution to the problem? I'm not saying I don't like it, I'm just curious where it came from and if we have any ability to "stay true" to the real-world facility.

Fechulo:
@erikquinn The airspace is based on Sectors 1 and 5 of the "San Juan Terminal Area":
sectors 1 5
This area is "normal-sized" while still maintaining some realism.
The whole document can be found [here](https://www.dropbox.com/s/fjnez9j3hmloqld/ZSU%207110.01A%20%281%29.pdf?dl=0).

erikquinn:
Oh wow, perfect. It actually looks really good then! Thanks for the verification.

Exfil:
I just tried out the version in http://zlsa.github.io/atc/ and there seems to be no arriving aircraft. Please check it out when you have time :)

Fechulo:
@Exfil It should be fixed with #664

Issue #576, opened 2016-06-13
SID text overlaps


erikquinn:
Whenever multiple SIDs share the same exit point, the SID name and exit point are written on the canvas at the location of that exit point, causing overlapping text. Logic could be added that would check to see if multiple things are trying to be written to the same location, and then have subsequent items offset above and/or below that position, resulting in a "new line" for each SID that ends at the shared exit point.

![image](https://cloud.githubusercontent.com/assets/5103735/16014857/b2723eb2-3158-11e6-8b51-b9cd682ddaec.png)

aldent95:
Got sick of this when trying to create Auckland International Airport (NZAA) so I went ahead and fixed it. Open to any suggestions to improve the fix.

![sids](https://cloud.githubusercontent.com/assets/8398954/17773489/d6ddff4c-65a2-11e6-858a-f62117c7bb7a.PNG)

eliuuk:
@aldent95 Looks great, I can't think of anything else to improve it though!

erikquinn:
Resolved, thanks to @aldent95's work on #643. Closing.

PR #643, opened 2016-08-18
Fixed issue #576 

aldent95:
Fixed issue #576 where the text for SIDS with the same transition would stack on top of each other.

Got sick of this when trying to setup NZAA as an airport so I went ahead and fixed it :)


eliuuk:
Great! This is perfect, especially with some of my airports where such SIDs overlap, like Delhi which is one of the worst offending ones. Is there anywhere we can see these changes in action? If so, that'd be great! 😉

aldent95:
@indianbhaji check my comment in the actual issue

erikquinn:
Beautiful, thank you! :smile: 

Merged via b0b093e.

PR #648, opened 2016-08-20
New Airport: KORD (Chicago O'Hare Intl.)

Fechulo:
The video map would not be there if it weren't for @erikquinn, so thank you :smile:! It seems like it's not too laggy if the fixes and SID's are disabled. (Although my computer is pretty beefy so others may experience more lag). You can take a look at it [here](https://fechulo.github.io/)!


eliuuk:
Perfectly fine for me, well done!

Fechulo:
@indianbhaji Thanks! Did you have any performance issues?

eliuuk:
@Fechulo Not at all, as much as I have a crap PC with only 3GB RAM, I've made it through so far! :+1:

Fechulo:
Great news then! Thanks for the feedback

PR #661, opened 2016-08-25
New Airport: KDFW (Dallas/Fort Worth Intl.)

Fechulo:
Another enormous american airport, except this time in Texas. As always, you can test it [here](https://fechulo.github.io/)!


erikquinn:
@Fechulo There's a lot going on with so many different video maps on... if you don't mind, let's reduce the video map to just arrival south and the final monitor maps. Will look like this:
image

That should eliminate some of the overlap, and reduce the draw load from 13k lines to about 6k lines.
[D10 RVM.txt](https://github.com/zlsa/atc/files/449774/D10.RVM.txt)

Fechulo:
@erikquinn Thanks for the video map, it was very difficult to remove the overlapping lines.

Ps: Accidentally closed the PR.

erikquinn:
[kdfw.json:923](https://github.com/Fechulo/atc/blob/fee4eb7529d8c28a2f75e90eb90d4b82f005bc53/assets/airports/kdfw.json#L923) `qfa/long` does not exist

Fechulo:
It does now! :smile:

erikquinn:
Amended to work with current version, and merged via d210989.

PR #664, opened 2016-08-30
Fix Transitions

Fechulo:


erikquinn:
:+1: Good call... I fixed it for SAWH, but forgot for TJSJ.

PR #665, opened 2016-08-30
[WIP] Flughafen München facelift

eliuuk:
München is Munich.
- [ ] Update STARs
- [ ] Revise traffic

If someone could assist with terrain, it'd be highly appreciated - my PC cannot handle the natural neighbour process.


Fechulo:
@indianbhaji Unfortunately, I tried creating the terrain for München in the past, but due to some of the limitations of the current method for generating terrain, a lot of errors come up. I'll try again, but I don't think I'll be able to get it to work.

eliuuk:
@Fechulo Alright, tell me the result, if it doesn't work, we'll just leave it. Thanks!

erikquinn:
Is this still being held?

Fechulo:
@erikquinn @indianbhaji To be honest, I haven't tried adding terrain since the last time, but I don't think I'll be able to make it work. The method we're using to generate the terrain works fine, but it's not perfect. Unfortunately, all of the imperfections of this method come up when trying to generate the terrain for EDDM.

eliuuk:
This is actually held because I haven't finished all the SIDs due to my poor prioritisation.

eliuuk:
Relocated to openscope/openscope#98, closing!

Issue #670, opened 2016-08-31
New Airport: Providence (KPVD)


Casper52250:
I would like to request that my home airport, KPVD, be added to the game. Whenever you get to it is fine. If there is anything that I can do to help the process along, let me know. I'm not a programmer with only basic intros to coding, but willing to learn (within reason).

Thanks for the great game and keep up the great work!

eliuuk:
Looks quite small, so shouldn't take too long. I haven't looked at any of the information about the airport in terms of navigation yet, so it might not as be tricky as the larger airports.

eliuuk:
I'll have a look at starting this now.

erikquinn:
@indianbhaji here's a video map for you as well. I put both the 5 and 23 flow maps up together.
[G90_RVM.txt](https://github.com/zlsa/atc/files/448520/G90_RVM.txt)
![image](https://cloud.githubusercontent.com/assets/5103735/18150730/236aa324-6fb7-11e6-9b92-6eafbbad38c2.png)

eliuuk:
@erikquinn Thank you very much for that!

Casper52250:
@erikquinn May I ask what the map shows? Is there something I should research to understand it?

Fechulo:
@Casper52250 The video map is what controllers in real life see. It shows fixes, waypoints, sectors, extended centerlines and other airports in the area. There's no need to learn anything, it's just a more aesthetically pleasing method of showing what the controller needs to see.

Casper52250:
I suppose it takes some getting used to, then... It seems a little tricky to pick out the names of the fixes/ waypoints among all of the other lines.

eliuuk:
You can disable the videomap at your own discretion. There is a button at the bottom which allows for this to happen.

erikquinn:
@casper52250 it differs greatly between facilities, but usually they will have some helpful markings depicting arrival and departure corridors, plus the important fixes and airspace boundaries. Not knowing all the sectors the facility is split up into can make it seem like there's lines all over the place, and it's just because we don't know what they're for. But we can sometimes find some documentation that explains the sectorization and markings.

erikquinn:
The ATC repository is being migrated to it's new home at [https://github.com/openscope/openscope](https://github.com/openscope/openscope), 
and thus, all issues are being closed. If this is still an issue with the latest version of the sim 
(accessible at [http://www.openscope.co](http://www.openscope.co)), or is a feature you still think we are lacking, 
[...]

PR #671, opened 2016-08-31
[ready] Theodore Francis Green Memorial State Airport, KPVD (request from @Casper52250)

eliuuk:
My first US airport. I don't particularly like doing US airports, as they require some scavenging (usually), but this one has been quite simple so far!

With 2 STARs and no SIDs, it's quite similar to EGNM, but with less traffic.

Requested by @Casper52250 in #670.

Thank you, @erikquinn for the videomap!


eliuuk:
Testing is available at https://indianbhaji.github.io/atc/

Fechulo:
@indianbhaji Due to the lack of SID's, when an aircraft is given the 'caf' command, an error occurs. The game doesn't crash, but it's still something worth noting. Looking at data from [FlightAware](https://flightaware.com/live/airport/KPVD/departures), The first waypoints for most aircraft are PUT, JUMPR, WOONS and PVD. Maybe you could add fake SID's with just a body. Similar to the LO, WZ and DK departures at UUDD.

eliuuk:
@Fechulo True, I'll just add SIDs like the ones in Brussels.

Issue #669, opened 2016-08-31
New Airports and Offline Play


Casper52250:
I'm sort of new here, and not (yet, anyway) a programming type, so a couple of questions:

It looks like there's a lot of work going on with the game, so are you guys taking requests for new airports? If not, is it hard to code a new airport (assuming it's not a big hub)? I'd like to see my home airport on the game, but I don't want to toss on the straw that breaks your backs.

Next, is there a way that the game could be played offline, or is it just based on the webpage itself? Is this something that can be done, or would this be a lot of work down a blind corridor?

I love the game, and I'm glad to see that there is work going into it to make it even better. If there's something a noob can do to help you guys out, let me know!

Fechulo:
@Casper52250 Hey, I'm glad you're enjoying the game!

The difficulty of adding a new airport depends mostly on the airport and the country it is in. Some airports have procedures, which can't be easily programmed into the game. Sometimes, the government of a country decides not to share any information on the airport which can make the process harder. Most international hubs are actually quite easy to program, they just take longer to make due to their size.

If you want, I can take a look at your home airport and tell you whether or not it's possible to add it to the game. If it is, I'd be more than willing to work on it.

There's currently no offline/desktop version of the game and I don't think there will be one any time soon. There's a lot of stuff that's being worked on at the moment and an offline version wouldn't be the number one priority. Then again, I'm not a coder, so I could be wrong.

If there's anything you'd like to know, I'll do my best to give you an answer (As long as I know what the answer is :smile:)

eliuuk:
Hello!

We do kindly accept requests for airports. All my airports are requested by a friend of mine who I introduced to the game, so it's normal for me. TJSJ in #642 by @Fechulo was requested by @Exfil was one of the most recent requests. If you want to request one, feel free to open up a separate issue for it!

**Playing offline**
I have not tried playing offline, so if this doesn't work, please tell me!

The game should be able to be played offline if you have TCPIP installed. This requires a third party application, such as [XAMPP](https://www.apachefriends.org/index.html) and [WampServer](http://www.wampserver.com/en/). Any program with Apache should be fine.

Download the code for this game [here](https://github.com/zlsa/atc/archive/gh-pages.zip) then extract it. If you're using XAMPP, move your folder to any location you wish and point XAMPP at it.

**Pointing XAMPP**
To point XAMPP at your game, open up the application and click 'Config' for for the row which says 'Apache'. Select the first option, 'Apache (httpd.conf)'. 
[...]

PR #674, opened 2016-09-01
Create offline.manifest (for offline play)

avdg:
This should enable offline play, which was being suggested in #669.

The files in the offline manifest are generated from static files. It won't include generated files. This means that those files won't be available offline. The game is playable offline though.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<html>
+<html manifest="offline.manifest">
   <head>
     <title>Air Traffic Control</title>
     <link rel="stylesheet" href="assets/style/main.css" />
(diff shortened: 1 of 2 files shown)


avdg:
Demo: https://avdg.github.io/atc/ (after files are loaded (in chrome it shows warning in console), cut internet connection, reload page)
Manifest generated: https://avdg.github.io/atc/offline.manifest

avdg:
Hmm.. seems that this feature should be overridden by service workers, although the manifest should probably still be used as fallback feature because of "old browsers" and given that [not all browsers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers#Browser_compatibility) have implemented this.

avdg:
I might be informed better about service workers after seeing this video: https://www.youtube.com/watch?v=cmGr0RszHc8

Still need to dig through a lot of documentation as the feature seems to be quite large.

Issue #683, opened 2016-10-03
[Airport Request] Kansai International Airport


Exfil:
Sorry I had to put it here in Issues but I don't know much about github or pull requests so I'll just leave this here hoping someone can move it to its appropriate place, thanks!

https://en.wikipedia.org/wiki/Kansai_International_Airport

eliuuk:
This would fit in well with the set of Japanese airports we have so far (Narita & Haneda)! I'll have a look tomorrow to see what I can excavate and develop on.

eliuuk:
I'll start work on this soon - I've got the information I need from the Japanese AIS!

erikquinn:
The ATC repository is being migrated to it's new home at [https://github.com/openscope/openscope](https://github.com/openscope/openscope), 
and thus, all issues are being closed. If this is still an issue with the latest version of the sim 
(accessible at [http://www.openscope.co](http://www.openscope.co)), or is a feature you still think we are lacking, 
please reopen the issue at the new repo.

*Please note that the vast majority of these issues have been copied to the new repository, or else are covered by other issues created there. See the below screenshot for what it looks like when your issue is known in the new repo:*

![image](https://cloud.githubusercontent.com/assets/5103735/22630816/5a32d794-ebce-11e6-9c46-d61a262e2b23.png)


 Thank you!


Closing this issue.

PR #696, opened 2016-10-16
[complete] RJBB, Osaka Kansai International Airport

eliuuk:
Osaka Kansai International Airport is just one of Japan's many airports. Requested in #683 by @Exfil!

Aviation is very important in Japan - after all, they are just a load of islands.


eliuuk:
Relocated to openscope/openscope#99, closing!

Issue #688, opened 2016-10-06
Halifax, NS (CYHZ) and Ottawa, ON (CYOW)


scott-j-smith:
Hi All,

As a beginner to the ATC sim I'd be happy to see more airports at a lower difficulty level.  Having grown up in Halifax, then lived in Ottawa for 20 years, I would love to see these added.  I think they would be good candidates for simpler airports.  

Halifax should be easier than Ottawa based on the fact Ottawa has a lot more regional airports around and I think some FL restrictions - but the amount of traffic is still not very high compared to many other airports already in the game.

https://en.wikipedia.org/wiki/Halifax_Stanfield_International_Airport

https://en.wikipedia.org/wiki/Ottawa_Macdonald%E2%80%93Cartier_International_Airport

Please consider for future additions.

Thanks,
Scott

eliuuk:
Definitely would be nice to see some Canadian airports added into the mix! I'll add that to my list.

scott-j-smith:
Thank you!

eliuuk:
Starting work on Halifax, expect that pull to be up soon!

scott-j-smith:
```
                                                                                  Can't wait!Thank you.                                                                                                                                                                                                                                                                                                                                         Sent from my BlackBerry 10 smartphone on the Bell network.                                                                                                                                                                                                                From: indianbhajiSent: Sunday, October 16, 2016 12:42 PMTo: zlsa/atcReply To: zlsa/atcCc: scott-j-smith; AuthorSubject: Re: [zlsa/atc] [Airport Request] Halifax, NS (CYHZ) and Ottawa, ON (CYOW) (#688)Starting work on Halifax, expect that pull to be up soon!
```

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or mute the thread.

[...]

PR #728, opened 2016-11-14
[startup] CYOW, Ottawa Macdonald-Cartier International Airport

eliuuk:
Ottawa International Airport is an airport (of course) that serves Ottawa, which is located close to the Canadian - US border.

Requested in #688!


eliuuk:
Relocated to openscope/openscope#102, closing!

